PR #3, opened 2015-08-07
Fixed a bug where files that require checkout are not found if you're…

vverbeek:
… using a path in the file.Src name. Because they are not found it will result in a 'File is not checked out' error.


erwinvanhunen:
Thank you! I'll merge this one.

PR #6, opened 2015-08-16
provision site group

ivanvagunin:
Moved PR from here https://github.com/OfficeDev/PnP/pull/1136

@erwinvanhunen , regarding new property for term group element. First I was thinking the same, but the problem with site collection term groups is that there is no straight way to create site collection term group using CSOM (and probably SSOM as well). The only available API is termStore.GetSiteCollectionGroup, where you can set createIfNotExist=true. This will create "default" site collection term group, called like "Site Collection - {site url}", and there is no way to change group name or id during creation. So making attribute like IsSiteCollectionGroup does not make much sense.
It may be a good idea to add special element inside TermGroups element (something like SiteCollectionTermGroup), but it requires schema and model changes. So I chose the simplest way, unfortunately it not that straight forward.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTermGroups.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTermGroups.cs
@@ -49,24 +49,37 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 TermGroup group = termStore.Groups.FirstOrDefault(g => g.Id == modelTermGroup.Id);
                 if (group == null)
                 {
-                    group = termStore.Groups.FirstOrDefault(g => g.Name == modelTermGroup.Name);
-
-                    if (group == null)
+                    if (modelTermGroup.Name == "Site Collection")
                     {
-                        if (modelTermGroup.Id == Guid.Empty)
-                        {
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Hi Ivan,

I investigated a bit on the server side, and indeed, to set a specific group as the site collection termgroup, we need a richer API than what is available to CSOM right now.

Alternatively, what we could do is still use the <TermGroup IsSiteCollectionTermGroup="True" /> construction, but then we need to skip the ID attribute, as that we cannot change that of an existing termgroup. The flow would then be (pseudo code):

var group = TermStore.GetSiteCollectionTermGroup(Site, true);
var group.Name = TermGroupName;
etc. etc.

ivanvagunin:
Hi Erwin,

That's true, but practically I don't see much need to change name of site collection term group (except following some naming convention). It may also be confusing, since GetSiteCollectionTermGroup(site, true) does not create a new group if existing group is renamed (as far as I understood it is tracked by ID saved to web properties). So adding several groups with different names to xml template will end up in renaming one group several times. 
Also I guess having one site collection group per site collection is more than enough, at least I don't see any case where you need have several of those. In our case it is important to have ability to create site collection level term set to be used in custom navigation control. Single site collection group works fine for that.

jansenbe:
Manually merged, thanks Ivan. Closing the PR now

PR #9, opened 2015-08-17
Added ability to specify url for the list view.

s-KaiNet:
When creating a view for a list through the CSOM it's not possible to provide `Url`. Usually SharePoint gets `Title` from `ViewCreationInformation` object and transforms it to url. The result is `All%20Items.aspx`. Things become even worse if I want view title in French `Tous les éléments` and url in English `AllItems.aspx`. View url becomes ugly `Tous%20les%20lments.aspx`. 
To workaround this, it is possible to setup title to be equal to url first (in order to force SharePont generate desired view url), then, when the view is added, we can update it with desired title.


jansenbe:
Thanks Sergei. Nice addition!

PR #15, opened 2015-08-21
Implemented usage of NetworkCredential.SecurePassword

jensotto:
Changed TimerJob class from using NetworkCredential.Password to NetworkCredential.SecurePassword to prevent using insecure string as storage for passwords.


jansenbe:
Thanks. Will have a look at this PR today

PR #16, opened 2015-08-21
Adding DefaultTermGroupSiteCollectionToken

mtringuyen:
Adding ParseString to ObjectExtensibilityProviders
ParseString two times on ObjectField to possible Parse DefaultTermGroupSiteCollection
Update TokenParser

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectExtensibilityProviders.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectExtensibilityProviders.cs
@@ -32,6 +32,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                     {
                         try
                         {
+                            provider.Configuration = parser.ParseString(provider.Configuration);
                             scope.LogInfo(CoreResources.Provisioning_ObjectHandlers_ExtensibilityProviders_Calling_extensibility_callout__0_, provider.Assembly);
                             _extManager.ExecuteExtensibilityCallOut(context, provider, template);
                         }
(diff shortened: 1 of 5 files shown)


erwinvanhunen:
Thank you for involvement and submission. Much appreciated! 

Can you give us some insight in what business case you are trying to solve with this pull request? Especially because in one of the changes you run the tokenparser twice, and we fail to understand why that would be needed. The token class you proposed is called DefaultTermGroupSiteCollectionToken, while the token actually is called termgroupsite. In which scenario would you want to use such a token?

mtringuyen:
I'm sorry for naming of the token not clear enough.
My scenario is I want provision a taxonomy field which use (ABC) term set in Default Term Group of site collection.
E.g.
<Name>TermSetId</Name><Value xmlns:q2="http://www.w3.org/2001/XMLSchema" p4:type="q2:string" xmlns:p4="http://www.w3.org/2001/XMLSchema-instance">{termsetid:Site Collection - {termgroupsite}:ABC}</Value>

Code changed call the tokenparser twice because first is for {termgroupsite}. The second is for parsing TermSetIdTokenthat we already developed.

Thanks,

erwinvanhunen:
Hi! Thank you so much for your explanation, much appreciated.

The request makes a lot of sense and we will credit you in our monthly call for your work. 
We however decided not to use your specific pull request, but implement it a bit differently. You can still nest the tokens, but the recursive calling will be done inside the tokenparser. E.g. you can now use, as you described in your scenario:

{termsetid:{sitecollectiontermgroupname}:ABC}

We named the tokens a bit different, and made available {sitecollectiontermgroupname} and {sitecollectiontermgroupid}

Check out the PR and it's implementation here: #21 

I will close this pull request.

PR #20, opened 2015-08-22
Improved SecurityExtensions methods to be called on any SecurableObject (i.e. Web, List, ListItem).

s-KaiNet:
Current implementation relies only on Web object to apply permission to. Some methods reworked to accept any SecurableObject (Web, List, ListItem).   
Note according breaking inheritance: caller should take care about breaking inheritance when make a call to AddPermissions....


jansenbe:
Nice work and good addition Sergei, thanks for this contribution! I'll merge this one still today before the september master merge

PR #22, opened 2015-08-24
Workaround for issue with IsUsingOfficeTheme (Issue #671)

lafe:
`IsUsingOfficeTheme` causes on SharePoint 2013 on premise an error, because the Theme Information cannot be found (error is like `"The object id "ThemeInfo-<GUID>" is invalid"`). This fix catches this exception and assumes that no Office Theme is used (in my experience, most companies use their own branding). This issue is also documented in the main repository as issue 671: https://github.com/OfficeDev/PnP/issues/671


jansenbe:
Thanks Lafe. Will have a look at this PR today

PR #33, opened 2015-08-30
Added required constructor for SP based Provisioning UX parts

wictorwilen:


PaoloPia:
Thank you Wictor, we appreciate your commitment and your efforts!

PR #49, opened 2015-09-04
Term groups provisioning: fix to search by group name with parsed tokens

Voyta:
Fix to first parse tokens in term group name, and check for existing group using parsed name.


erwinvanhunen:
Thank you for your submission!

PR #50, opened 2015-09-04
Change to log fieldId, not field.Id as field.Id is not initialized at this point

Voyta:
Existing code was failing on adding fields to content type


erwinvanhunen:
Very good catch! will merge

PR #73, opened 2015-09-15
some provisioning engine fixes

ivanvagunin:
Some early versions of CSOM for SP2013 on-premises does not support MajorVersionLimit, so it breaks provisioning if versioning enabled,


ivanvagunin:
Added two more fixes:
1) ObjectTermGroups.ExtractObjects crashes on-premises if only default keywords term store is set (no default site collection term store set)
2) SecurableObjectExtensions.ReplaceGroupTokens crashes when there are no associated groups, but it still tries to replace the tokens (even you don't use tokens)

jansenbe:
Thanks Ivan!

PR #74, opened 2015-09-16
Implemented fix for bug #72

paulpascha:
I've implemented a fix for bug #72


jansenbe:
Hi Paul, thanks for this fix. I've taken your approach but used a different implementation, hence I'm closing this PR...but the change is in.

PR #76, opened 2015-09-20
Expose search export for all levels to string

afsandeberg:


jansenbe:
Thanks for this contribution Mårten. After evaluation we decided to not merge this one as obtaining search settings via a clientcontext extension is confusing + we extension methods on site and web level to the same

PR #81, opened 2015-09-22
Views are using wrong name

lafe:
Up until now, created views might use a wrong file name. If the original was called "All Items" and the file was "AllItems.aspx", the restored view was restored as "All Items.aspx" using the display name of the view.
This caused some issues in our environment, so I changed the behavior to extract the file name from the URL. This file name is used as the first display name of the view and is therefore used as the file name. Afterwards, the display name will be changed to the correct file name.
If an error should occur, the behaviour switches back to the current behaviour by using the display name as the view and file name.


s-KaiNet:
I've already done this in [here](https://github.com/OfficeDev/PnP-Sites-Core/pull/9), but it's not still merged...

jansenbe:
Thanks lafe...needed addition, but closing as similar implementation done by @s-KaiNet was merged.

PR #85, opened 2015-09-25
Added option to load extra properties on list extension GetListByTitle and support for lookup fields, which isn't a site column.

SchauDK:
When using list extension GetListByTitle, I often have a need for loading extra properties.
Eg. 
    var list = web.GetListByTitle("ListTitle", l => l.ContentTypes);

This small change will make it possible to do so while keeping backward compability.


SchauDK:
Export has become really slow because Get Security is called extremely many times.
I've fixed this bug by moving the GetSecurty method outside the FieldCollection loop within the ListInstance object handler.

PR #86, opened 2015-09-27
BugFix for TimerJob framework to run On-Premises using AppOnly authen…

suhasrs:
The existing Remote timer job framework does not work for On-Premises AppOnly authentication scenario (without entering Network credentials). This bugfix is for fixing this scenario.


jansenbe:
Hi suhasrs,

Thanks for your contribution. We've evaluated this PR and decided not to merge it because app-only is supported in the framework using low trust (ACS). You're introducing high trust (S2S) which is something we're not really working on in PnP. Would recommend to add these changes in your own codebase.

PR #93, opened 2015-10-05
Making sure IDs are loaded for all FieldLinks in a ContentType with EnsureProperties

olemp:
I created a issue earlier today (https://github.com/OfficeDev/PnP-Sites-Core/issues/91) related to Content Types and Fields. I got a error when adding certain fields to a Content Type. The ID of the field was not loaded properly.

I went through and debugged the ContentType-handler and found that sometimes the fields wasn't loaded properly.

With this pull request I'm making sure IDs are loaded for all FieldLinks in a ContentType with the override EnsureProperties().

### DIFF
--- a/Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
+++ b/Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
@@ -739,6 +739,10 @@ namespace Microsoft.SharePoint.Client
         public static void AddFieldToContentType(this Web web, ContentType contentType, Field field, bool required = false, bool hidden = false)
         {
             contentType.EnsureProperties(c => c.Id, c => c.FieldLinks, c => c.SchemaXml);
+            foreach(FieldLink fieldLink in contentType.FieldLinks)
+            {
+                fieldLink.EnsureProperties(fl => fl.Id);
+            }
             field.EnsureProperties(f => f.Id, f => f.SchemaXml);
 
             Log.Info(Constants.LOGGING_SOURCE, CoreResources.FieldAndContentTypeExtensions_AddField0ToContentType1, field.Id, contentType.Id);
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs
@@ -102,8 +102,13 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 
                         if (checkedOut)
                         {
-                            targetFile.CheckIn("", CheckinType.MajorCheckIn);
-                            web.Context.ExecuteQueryRetry();
+                            try {
+                                targetFile.CheckIn("", CheckinType.MajorCheckIn);
+                                web.Context.ExecuteQueryRetry();
+                            } catch(ServerException e)
+                            {
+                                // TODO - File could not be checked in
+                            }
                         }
 
                         // Don't set security when nothing is defined. This otherwise breaks on files set outside of a list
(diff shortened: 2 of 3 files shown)


olemp:
Tried to create another PR, but https://github.com/olemp/PnP-Sites-Core/commit/de7389b36e9bbc839a58adfde9be511f8bf7b3e5 was merged into this one.

Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
```diff
@@ -739,6 +739,10 @@ public static void AddFieldToContentTypeByName(this Web web, string contentTypeN
         public static void AddFieldToContentType(this Web web, ContentType contentType, Field field, bool required = false, bool hidden = false)
         {
             contentType.EnsureProperties(c => c.Id, c => c.FieldLinks, c => c.SchemaXml);
```
rgylesbedford: changing line 741 to be the following would be a cleaner fix:
contentType.EnsureProperties(c => c.Id, c => c.SchemaXml, c => c.FieldLinks, c => c.FieldLinks.Include(fl => fl.Id, fl => fl.Required, fl => fl.Hidden));

SchauDK: It would be cleaner if EnsureProperties supported Include, but it doesn't.
In my opinion, in this case we're better off not using EnsureProperties at all.

olemp:
@SchauDK How do you want to make sure the properties are loaded then?

Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
olemp: @SchauDK Would you properly load the Content Type, Field and FieldLinks instead?

s-KaiNet: In your case it's better to use following: 

```
contentType.EnsureProperties(c => c.Id, c => c.FieldLinks, c => c.SchemaXml);  
contentType.Context.Load(contentType.FieldLinks, w => w.Include(fl => fl.Id, fl => fl.Required, fl => fl.Hidden));
contentType.Context.ExecuteQueryRetry();
```

olemp: @s-KaiNet I think that's a good approach.

SchauDK: @olemp @s-KaiNet  Yep, that's how I'd do it.
When I get a little time I'll look into if it's possible to extend EnsureProperties to support Include. That would be nice ;-)

s-KaiNet: @SchauDK, I will look today's evening (GMT +3) :)

olemp: @SchauDK That would be very nice indeed - but I think the whole procedure could be done differently. Shouldn't be neccessary to load the FieldLinks for the content type for each file you want to add.

PieterVeenstra: Related to this fix. I’m using the latest code and I found that when I’m adding 2 or more fields to an existing content type the following line in Field and ContenttYpeExtensions.cs ( line 748) still returns the same id exception

  var flink = contentType.FieldLinks.FirstOrDefault(fld => fld.Id == field.Id);

The first field is added but the 2nd field can’t be added.

s-KaiNet:
Created #99

VesaJuvonen:
Thanks for this. This does seem to fix the whole thing. We'll get this included during the weekend after second verification. Especially the field reference issue is causing wider headache as well, so we'll most likely push update directly to the Oct Nuget release to get this addressed immediately.

Will wait for second confirmation on the fix and we'll get things then included to avoid surprises.

PaoloPia:
Hi! Thanks for your contribution, we really appreciate your commitment. Unfortunately, this PR targets the master branch, instead of the dev branch, and I need to close it. However, thanks to your feedbacks and to PR #99, which has already been merged, we will now fix the engine in the dev branch and will release an updated Oct Nuget package as soon as possible, as stated by Vesa.

VesaJuvonen:
We'll absolutely ensure thought that all of the people related on this pull requests will get credits on reporting these findings and providing solution for them regardless that things were actually manually merged to dev branch. 

Thank you for reporting these. Community programs absolutely need assistance from the community for ensuring that we can provide value for each other.

PR #94, opened 2015-10-05
Getavailabepagelayoutfix

kendomen:
ctx.Web.GetProvisioningTemplate() throws {"Data at the root level is invalid. Line 1, position 1."}

I added a check in GetAvaliablePageLayouts(Web web) in ObjectPublishing.cs to handle this error:

// if (defaultLayoutXml != null)
if (defaultLayoutXml != null && defaultLayoutXml != "__inherit")
{
defaultPageLayoutUrl = XElement.Parse(defaultLayoutXml).Attribute("url").Value;
}


kendomen:
Thanks!

erwinvanhunen:
Thank you, Ken! Good catch.

PR #95, opened 2015-10-07
fix for AddFieldToContentType does not set Required=false for field link

ivanvagunin:
FieldAndContentTypeExtensions.AddFieldToContentType does not apply Required=false when adding field link to content type. It ends up in that in situation when provisioning engine add fields marked as required even when you explicitly set Required attribute to false in pnp:FieldRef node.

### DIFF
--- a/Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
+++ b/Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
@@ -760,8 +760,15 @@ namespace Microsoft.SharePoint.Client
                 flink = contentType.FieldLinks.GetById(field.Id);
             }
 
-            if (required || hidden)
-            {
+			//update field link required and hidden properties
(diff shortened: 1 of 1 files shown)


Core/OfficeDevPnP.Core/AppModelExtensions/FieldAndContentTypeExtensions.cs
```diff
@@ -760,8 +760,15 @@ public static void AddFieldToContentType(this Web web, ContentType contentType,
...
             }
 
-            if (required || hidden)
-            {
+			//update field link required and hidden properties
+			if (!flink.IsObjectPropertyInstantiated("Required") || !flink.IsObjectPropertyInstantiated("Hidden"))
+			{
+				web.Context.Load(flink, f => f.Required, f => f.Hidden);
+				web.Context.ExecuteQueryRetry();
+			}
```
s-KaiNet: This can be simplified with: 

```
flink.EnsureProperties(f => f.Required, f => f.Hidden);
```

PaoloPia:
@ivanvagunin thanks for your contribution. As usual, we really appreciate your commitment. And thanks to @s-KaiNet for the suggested simplification.

PR #97, opened 2015-10-08
Added ReIndexList, method to queue list for reindexing

afsandeberg:


afsandeberg:
Added a small fix for for setting propertybag value on list.
When using a list object from ClientContex.Web.GetListByUrl the extension methods SetPropertyBagValue methods doesn't (the values don't stick).

afsandeberg:
Closed, more files than intended was pulled

PR #100, opened 2015-10-09
Fixed exception thrown when trying to assign same permission twice. 

s-KaiNet:
This code:
```
clientContext.Web.AddPermissionLevelToGroup(_testGroupName, RoleType.Contributor, true);
clientContext.Web.AddPermissionLevelToGroup(_testGroupName, RoleType.Contributor, false);
```
throws an exception `The permission level specified is already added to the collection` in current pnp version. But sometimes in complex scenarios it is possible that code tries to set same permission twice. This was fixed. Also simplified methods `AddPermissionLevelImplementation` and `RemovePermissionLevelImplementation`


jansenbe:
Thanks Sergei. I'll have a look at this one and process it.

PR #105, opened 2015-10-11
Added option to load extra properties in GetListByTitle

SchauDK:
Also fixed problem with using Equals in LoadQuery.


jansenbe:
Hi Michael,

Thanks for this PR. The idea you're proposing is definitely interesting and could be a way for Core library users to optimize their code. Before we can merge this we need evaluate this approach for the whole core library: we either add this option everywhere (when possible) or not at all. If we only add it for a few methods this will result in inconsistent and confusing behavior.

To further track this I've created issue #155

This also implies that we're closing this PR now. If we later on decide to roll out this approach in the core library you'll receive credits for the idea.

SchauDK:
Thanks, I can only encourage that it will be implemented throughout the framework ;-)

PR #107, opened 2015-10-11
ReIndexList and fix for list SetPropertyBagValue

afsandeberg:
ReIndexList a method to queue list for reindexing

And a small fix for for setting propertybag value on list (rootfolder).
When using a list object from ClientContex.Web.GetListByUrl the extension methods SetPropertyBagValue methods doesn't work (the values don't stick).


jansenbe:
Thanks Mårten. I'll have a look at this PR and process it

jansenbe:
Merged, but somehow github is not closing this one...so closing manually

PR #112, opened 2015-10-12
MajorVersionLimit is not supported on-premises causes runtime exceptions

okms:
MajorVersionLimit is not supported on-premises and checking this property will cause a serverside error resulting in a runtime exeption. Adding compiler flags to exclude checking these properties in V15 builds

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -617,11 +617,13 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                         existingList.EnableVersioning = templateList.EnableVersioning;
                         isDirty = true;
                     }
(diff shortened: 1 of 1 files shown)


okms:
If the change looks familiar its because I tried piggybacking this change onto pull-request #93 in commit https://github.com/olemp/PnP-Sites-Core/commit/f16b619cd56039db27fbda1f93bb0d99b9dc9093 

Since #93 is closed, im submitting this change as a separate request.

ivanvagunin:
Hi,
In version 15 of CSOM MajorVersionLimit is not supported only by server side, so there is no need to exclude code from compiling by using directives. Checking IsObjectPropertyInstantiated before accessing property value should be enough to avoid runtime crash. 
UpdateList works for me on-premises with Sep 2014 CU and later, CreateList can be fixed the same way.
I would propose to use that approach just in case they fix it in some further CU (man can dream ))


Br, Ivan

erwinvanhunen:
Thank you for your submission! We implemented your changes manually, due to conflicts with existing code. Obviously you will receive credits for this change.

okms:
thank you @erwinvanhunen :+1:

PR #119, opened 2015-10-15
Added support for lookup fields, which isn't a site column

SchauDK:
When a lookup field is created/update the List attribute is temporarily removed.
In this case the field cannot be created if RelationshipDeleteBehavior is set.
I've fixed this issue and is also making sure Indexed attribute is set to True if RelationshipDeleteBehavior is either Cascade or Restrict.


erwinvanhunen:
Thank you for your submission! Much appreciated.

Issue #104, opened 2015-10-11
CT check for not existing site column


braegelno5:
On SP Prem it is possible to rollout a wsp with a ct definition with one ore more not existing linked fields. This bug comes only on light if you use this ct..
So if we export the ct to our xml we need a check for really existent fields. On the CSOM import side we get big trouble..
Tomorrow if I have access to my production environment I will upload a patch for this..

PaoloPia:
Hi Torsten, Is this issue still open, even after the changes we made in the latest months? Thanks.

braegelno5:
if i remember this correct, there was also a pr for this.. anyway i have changed the known wsp on our side, where i detected this behavior. On the other side we should have such a check in the engine, and the engine should only write an error to the log for the not existing objects and never stop there work.

PaoloPia:
Thanks Torsten, we usually prefer to block the provisioning in case of any error, instead of proceeding making the user making wrong assumptions. I mean: if we can't properly provision the template, we prefer to stop the provisioning and make the user aware of it, rather than logging a line an proceeding with a potentially inconsistent result. I will close this issue, because there should be a merged PR for this. If you will experience it in the future, I will be more than happy to re-open it. Thanks again.

braegelno5:
not so hurry italian man;) i have make a new pr (#401), and this will throw an error with field id and text, so the enduser using powershell will be able to find the bug inside the template xml.

PaoloPia:
Thanks for your contribution. I will double-check the PR and let you know.

PR #125, opened 2015-10-17
all is assigend to work properly on prem sp and solves also issue #104 && #4

braegelno5:


PaoloPia:
Hi Torsten,
We really appreciate your contribution and your commitment. Sorry for the delayed processing of this PR. Unfortunately we cannot merge this, because there are some conflicting changes that we cannot include. Could you be so kind to split your changes into multiple PRs (one for each specific target), so that it will be possible for us to merge them one by one. Thank you again for your help. Thus, I will close this PR.

PR #127, opened 2015-10-20
ObjectHandlerBase.Tokenize 'url.IndexOf' failing on encoded Uri

Arceau:
String comparison was failing due to Uri r encoding webUrl, this change decodes the PathAndQuery so that url.IndexOf is successful if there are any escaped characters.


erwinvanhunen:
Very good catch! Thank you!

PR #132, opened 2015-10-22
Fixed reintroduced bug where tokens should be excluded when parsing f…

SchauDK:
…ield.

Parsing of fields was added in following commit.
https://github.com/OfficeDev/PnP-Sites-Core/commit/c9a55b6016f052bcdc9798f460661b6df959d6db

It reintroduced the issue with ~sitecollection token being parsed. 
It means that eg. JSLink won't work as it needs the ~sitecollection token and not the parsed value.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectField.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectField.cs
@@ -117,7 +117,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                     {
                         existingFieldElement.Attributes("Version").Remove();
                     }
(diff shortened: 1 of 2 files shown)


rgylesbedford:
Hi Michael, is it possible to add some tests to ensure that if this regression happens again, the tests fail?

 * [ObjectFieldTests.cs](https://github.com/OfficeDev/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectFieldTests.cs)
 * [ObjectListInstanceTests.cs](https://github.com/OfficeDev/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectListInstanceTests.cs)

erwinvanhunen:
Good catch. Will merge

SchauDK:
@rgylesbedford  Sure, I'll add some tests

PR #134, opened 2015-10-22
xinclude fallback support and recursive parsing

ivanvagunin:
Add support for XInclude/Fallback element and recursive XInclude parsing


PaoloPia:
Thanks for your contribution. As always, we really appreciate your commitment!

PR #141, opened 2015-10-28
Added skipping of tokens ~site and ~sitecollection 

bjdekker:
When creating a field containing a JSLink in an XML provisioning template, the ~sitecollection token is incorrecty replaced with the actual url of the site collection, this way the display template is not loaded. I added the skiptokens ~sitecollection and ~site to the call of parsing function within the createfield and updatefield functions.


rgylesbedford:
[Pull Request 132] (https://github.com/OfficeDev/PnP-Sites-Core/pull/132) fixes this same issue for both list fields and site fields.

bjdekker:
You're correct, didn't see that one, but it fixes the same problem in the same qay. I will close my pull request. Thnx!

PR #144, opened 2015-10-29
Removed missing app config file from csproj.

yohanb:
Here goes for my first pull request to PnP!

The *app.config* file in the test project is references on the *csproj* but isn't physically there in the repository.  I removed the reference to the file in the *csproj*.

Cheers,

Yohan


jansenbe:
Thanks Yohan. 
The missing app.config is by design: we need this file to run the unit tests, but since we don't people to check-in the file (it can contain credentials) it's excluded from source control. The steps to follow are copying the app.config.sample and rename it to app.config.

So I'm not going to merge this pr as it would break our unit testing.

PR #158, opened 2015-11-03
XMLPnPSchemaV201508Formatter - Fix to not to serialize empty ComposedLook

Voyta:
Fix for the following issue:

- Create new template (automatically creates empty instance of ComposedLook)
- Do not set ComposedLook
- Save template to XML file - will create ComposedLook node with empty attributes
- Try to read template from resulting XML file

Current fix compares current composed look with empty one, and saves it only if it is not empty.


PaoloPia:
Thanks for your submissions, we appreciate your commitment. Your PR has been merged!

PR #160, opened 2015-11-03
Serialize object Security only if it is not empty

Voyta:
Serialize object Security only if it is not empty (File.Security is automatically initialized to empty ObjectSecurity object. When template is applied (ObjectFiles) code does check if security object is not empty and only then applies. But on serialization it did not check and produced empty Security nodes in XML files )


PaoloPia:
Thanks for your submissions, it has been merged. We appreciate your commitment.

PR #161, opened 2015-11-03
Rethrow provider exceptions

afsandeberg:
When using/developing extensibility providers it’s difficult to follow the execution when exceptions inside the providers isn’t exposed. You could use the trace log capabilities but that give you a little bit too much, I just what the error.

Suggestion:
- Log the exception as an error
- Rethrow the exception


PaoloPia:
Thanks for your contribution, it has just been merged!

Issue #177, opened 2015-11-10
UserCustomAction ignores sequence on ScriptLink


Trickass:
When creating a CustomAction with location *ScriptLink*, the *Sequence* property is ignored.
Could you please change this?

The relevant code is located in *AddCustomActionImplementation* method in *NavigationExtensions.cs*.

PaoloPia:
Hi, 
Thanks for your feedback. This one should be fixed. I will close the issue.

PR #186, opened 2015-11-13
Do not ignore Sequence property for scriptlink type UserCustomAction

Trickass:
Moved "Sequence" property setting out of UserCustomAction type checking.

Sequence is an important property to be used when adding multiple ScriptLink custom actions which might depend on eachother.

Issue reference: #177 

Edit: additional change in pull request #188 by @luismanez. That one seems to be preferred solution since it incorporates change to provisioning engine as well.


erwinvanhunen:
closing. Manually merged through PR #189

PR #197, opened 2015-11-16
Composed look null values

Arceau:
This request revolves around the fact that when a new PWA project is created the "Current" composed look has null values for all of the fields, this request basically introduces some extra checks to ensure values are not null before being used as a comparison.


jansenbe:
Similar capability has been added via other merged PR. Please let us know if you can still repro the issue. I'll close this PR now.

PR #199, opened 2015-11-18
Changes to GetCurrentComposedLook method

tavikukko:
Changes to GetCurrentComposedLook method, to work also in sites are not in english. In English sites the current name of the theme is "Current" but in finnish sites for example it is "Nykyinen", so the method will fail. But if we use the id of the item, which is always 1, then we will get always the right theme.


tavikukko:
pushed to wrong branch. made new one in dev branch.

PR #207, opened 2015-11-23
Changes to GetCurrentComposedLook method

tavikukko:
Changes to GetCurrentComposedLook method, to work also in sites not in english. In English sites the current name of the theme is "Current" but in finnish sites ex. its "Nykyinen", so the method will fail. But if we use the id of the item, which is always 1, then we will get always the right theme.


jansenbe:
Thanks Tomi. You're 100% right on the need to fix this but I've taken another approach because in my testing I've seen cases where the current item is the item with item id = 1. I've now simply used the server resource files to acquire the proper name for "current".

tavikukko:
Hi Bert!
That's great approach, I did not know it was possible to use server resource files with CSOM. Now I do, thanks!

PR #226, opened 2015-12-02
fix for provisioning engine pnp:Page parsing

ivanvagunin:
Page element in provisioning template is parsed in a way that Security property is not null, even if Security element for the page is not specified. That brings trouble with permissions removal for the page, when Security element is omitted.


jansenbe:
Thanks Ivan

rfjschouten:
This issue also applies to ListInstance datarows added. They'll end in a state with removed permissions too, when no security element is specified.

PR #240, opened 2015-12-09
Implemented conditional processing of object handlers

erwinvanhunen:


SchauDK:
Excelent change. Really useful.

PR #258, opened 2015-12-16
Added paged property to viewCreationInformation

Laskewitz:
Fix for PowerShell issue 174: https://github.com/OfficeDev/PnP-PowerShell/issues/174


erwinvanhunen:
Hi! Thank you for your submission! There is however an issue with the way you added the optional Paged variable in the method signature.

You change it from

```csharp
    public static View CreateView(this List list,
                                      string viewName,
                                      ViewType viewType,
                                      string[] viewFields,
                                      uint rowLimit,
                                      bool setAsDefault,
                                      string query = null,
                                      bool personal = false)
````

to

```csharp
    public static View CreateView(this List list,
                                      string viewName,
                                      ViewType viewType,
                                      string[] viewFields,
                                      uint rowLimit,
                                      bool setAsDefault,
                                      string query = null,
                                      bool paged = false,
                                      bool personal = false)
```

Given the existing method signature, I could write my call to the method like this

```
list.CreateView("Demo View",30,ViewType.None, 30,true,null,true);
```

The last true in the method call would mean that I'm creating a personal view.

If I would merge your PR, the call to the method would suddenly mean that I create a shared view with paging enabled. That is a breaking change and we unfortunately can't merge your PR of this. The rule of thumb is that when you add default parameters to a method signature, you have to add them to the end of the method signature.If you fix your PR I can merge it after evaluation.

Thanks!

PR #304, opened 2016-01-11
Extend role assignment extraction process to deeper level

LiyeXu:
Extend role assignment extraction process to even deeper level so that resources with unique permission at those levels could be reached as well.


jansenbe:
Thanks Eric, will look into this.

jansenbe:
Merged :-)

PR #324, opened 2016-01-30
Added method BindDocumentTemplate

nanddeepn:
Added method BindDocumentTemplate, which Binds document template to content type.


jansenbe:
Hi Nanddeep,

Thanks for this pull request. After evaluation in the core team we decided to not merge this PR as it's essentially only setting a single property. We would recommend you to add this extension method to your project code so that you can still use it in your projects.

PR #335, opened 2016-02-05
PageExtension AddWikiPage impovements

sven-pohl:
PageExtionsions.cs 
  AddWikiPage 
- Removed hard coded URL path "sitepages"
- Return of server relative URL instead list relative URL
- Return site relative URL if page already exists


sven-pohl:
Wron branch selected

PR #372, opened 2016-02-13
Added some useful methods to FileCollection.

paulpascha:
AddFromResourcesFolder allows adding files recursively from the specified resources folder. Very useful for provisioning files to Style Library for example.

RemoveUnmodifiedSinceDate removes all files from the collection which have not been modified since the specified date. Excluding files that haven't been modified to increase Provisioning performance.


PaoloPia:
Hi Paul, I've just added an issue in the PnP Provisioning Schema repo. In order to properly support this capability we need to update the provisioning schema. We will do that in the next major version of the schema. Thanks for your suggestion. In the meantime I will close this PR and I will keep you updated.

PR #391, opened 2016-02-17
Modified SharePoint connector to use web root folder rather than list folder when container is not supplied 

skarta:
Modified SharePoint connector to use web root folder rather than list folder when container is not supplied.

Addresses issue when generating a template using the provisioning engine from a site where the welcome page is for example default.aspx in site root folder rather than a page in a list.


jansenbe:
Thx skarta, good addition

PR #400, opened 2016-02-21
Fixed issue #227.

PaoloPia:


humz00:
Hi  @PaoloPia ,
I am still receiving this error when re-applying a provisioning template to a site where an events list (106) already exists. 

The issue does not occur if i remove the calendar list and then re-apply the template.

I am using version 2.3.1604.0 of the officeDevPnp.Core.Online library.

Thanks,

oyvindam:
This is still an issue on 2.10.1612.0

PR #401, opened 2016-02-21
fix for #104 site column not exists on adding to existing content type

braegelno5:
fix with check for existing field, if not - log for debug and throw an error, with the same description like the debug log, without description the bug in the pnp:templatexml is not findable for the end user, using powershell or any adapted sample code


PaoloPia:
Hi Torsten,
Unfortunately in this PR we have almost the same issue of the previous one (PR#356). I cannot do proper testing or solve the conflicts without the source branch. Please re-submit your PR keeping the source branch alive. If you have any issue keeping the source branch alive, please let us know and we will be more than happy to support you in managing the task. We're afraid about not being able to merge this, but I will close this PR, waiting for a new one. Thank you again.

Issue #429, opened 2016-02-29
Importing workflows creates multiple instances of the same workflow


PieterVeenstra:
I've exported site with list workflow. Then I imported this workflow and I'm getting multiple instances of the exported workflow.

Every time I run the following command I get another instance of the same SPD workflow:

Apply-SPOProvisioningTemplate C:\TFS\Installer\Data\WebTemplates\Home\template.xml.updated -Handlers Workflows

m-carter1:
This looks like the same issue here #323

We have a fix almost ready to submit

PaoloPia:
This one has been closed by PR #436 .

Issue #323, opened 2016-01-29
Duplicated Workflow Subscription after reapplying provisioning template


mhzawadzki:
A simple subsite1 with a simple SP Designer 2013 workflow associated with a List1, list has default local content type - Item. Get a provisioning template of this subsite1 and Apply this template to a different site collection's subsite2. You can run the workflow on subsite2 List1 with no issues.

Run the provisioning process again against the same subsite2 and use the same template as before. This time workflow subscription was added again (incorrect), but workflow definition wasn't (correct, as there was no delta change). This leads to the same workflow starting twice on an item in List1 on subsite2.

When I Get a new provisioning template from subsite2 I see duplicated WorkflowSubscription tag with the same DefinitionId:

<pnp:WorkflowDefinition Id="ad9d6f7c-69b5-47d7-9b4a-7b0dad63c098"
<pnp:WorkflowSubscription DefinitionId="ad9d6f7c-69b5-47d7-9b4a-7b0dad63c098"
<pnp:WorkflowSubscription DefinitionId="ad9d6f7c-69b5-47d7-9b4a-7b0dad63c098"

This looks like a bug, as there shouldn't be any duplication for workflow, as there was no change on either of subsites.

asumner:
I wonder if delta functionality has been implemented on workflows yet.

m-carter1:
Also (after provisioning) if the associated workflow on the list is edited and re-published it will create a duplicate workflow subscription.

I've taken a quick look at the provisioning code for Workflows and I noticed that for the Subscription PropertyDefinitions it only takes the TaskListId and HistoryListId 
(line 230 at OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.ObjectWorkflows)
`foreach (var p in subscription.PropertyDefinitions
                            .Where(d => d.Key == "TaskListId" || d.Key == "HistoryListId"))
                        {
                            workflowSubscription.SetProperty(p.Key, parser.ParseString(p.Value));
                        }`

If I comment out the where part and allow it to copy all of the PropertyDefinitions for the new subscription it seems to solve this issue, in that it will no longer re-add the workflow (gives duplicate error message) during provisioning and editing & re-publishing the list workflow also no longer creates a duplicate!

Is there any reason not to copy over all Property Definitions? 

I suspect there's a particular property (or a few) that are required for the fix but I'm unsure which ones...

asumner:
[...]

PR #431, opened 2016-02-29
Fixes for Workflow duplication issues #323 and #429

m-carter1:
Fixes for Workflow duplication issues #323 and #429

Added a check for existing Workflow Definitions and Subscriptions before adding them to stop duplication.
Included Subscription.Id and Subscription.Name property definitions when creating the Workflow Subscription as this fixes an issue where the workflow subscription is duplicated if it is re-published in SP Designer.


PaoloPia:
Hi Matt,
Thanks for your great contribution. I have just merged your changes together with some other fixes within PR #436. Your commits have been kept so that you will be fully credited. I will close this one, because it is included in the other PR. Thanks again!

Issue #230, opened 2015-12-05
Calculated list column fails at creation


PaoloPia:
Please refer to this issue: https://github.com/OfficeDev/PnP-Provisioning-Schema/issues/52

PaoloPia:
As far as I know this issue has been fixed. I will close it, but feel free to re-open it in case of any further need, or if you will experience any other trouble while using calculated fields. Thanks.

PR #432, opened 2016-03-01
Fix for Calculated columns #230

m-carter1:
Fix for https://github.com/OfficeDev/PnP-Sites-Core/issues/230

When extracting calculated fields, the formula for in the SchemaXml is using the Internal Field Names which when provisioning a site will throw an error as it needs the Display Names.
This fix gets the Formula from the Field, which contains the Display Names and replaces the Formula in the SchemaXml before it adds the field.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -1329,9 +1329,25 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 }
                 else
                 {
-                    var schemaXml = ParseFieldSchema(field.SchemaXml, lists);
+                    //var schemaXml = ParseFieldSchema(field.SchemaXml, lists); //not being used..
                     var fieldElement = XElement.Parse(field.SchemaXml);
                     var listId = fieldElement.Attribute("List") != null ? fieldElement.Attribute("List").Value : null;
+
+                    //get field type
+                    var fieldType = fieldElement.Attribute("Type") != null ? fieldElement.Attribute("Type").Value : null;
+
+                    if (fieldType != null && fieldType == "Calculated")
+                    {
+                        var calcualtedField = (FieldCalculated)field;
+
+                        // update SchemaXml value with field formula value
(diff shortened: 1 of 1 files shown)


braegelno5:
Here we should check, if a site with language en, it and de for example has driven a template in german context, so the formula has german dispay names; now we roll out this template to a site with en, it, de and spanish language, but the primary language is italy. So the german formula will not throw an exception?

PaoloPia:
Hi Matt and Torsten,
I think we should tokenize the fields in the formula, in order to support the proper display name, based on the site language. Let me do that: I will sync with Erwin (who is our tokenizer master) and we will provide a suitable solution that works with any language. The only potential issue that I see, is that probably we will not be able to complete the task before the next master merge (which will occur tomorrow afternoon). Nevertheless, we will start from here. Thus, you will be both credited for your contribution. I will keep this PR open until the solution will be ready. Thanks!

PaoloPia:
This issue should have been fixed by this PR: https://github.com/OfficeDev/PnP-Sites-Core/pull/466.
Thus, I will close it.

m-carter1:
Hi @erwinvanhunen , this issue doesn't seem to have been resolved.

The tokenizer looks great but the generated schemaXml isn't being used
https://github.com/OfficeDev/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs#L1407

Simple fix for this is to use the generated xml which is fine but there's another issue where the tokens in the field formulas aren't being swapped out so it will fail to create the fields.

It doesn't look like the tokens are being created anywhere for the list fields and I'm not sure how you can create the tokens before you create the field, and you can't create the field with the tokens still in the formula?

thanks, Matt

jensotto:
Hi, I just want to let you know that this implementation introduces a major bug. If applying the provisioning template twice the calculated field will not work anymore. 
I guess this is because of some internal SharePoint manipulation when changing the SchemaXML note that the provisioning template contains the field title (display name), but when provisioned to SharePoint, SharePoint translates it into the internal name and adds field references for each field. When applying the template for the second time it appears that SharePoint does not perform the translation of field title to field internal name. 
The formula looks like this after second deploy when I edit the field in the browser: **=#NAVN?**

jansenbe:
Fixed using #817

PR #434, opened 2016-03-03
Added SetJsLinkCustomizations site and list fields

thomasberman:
Added a Field extension method to add jsLink to field.
Added a List extension method to add jsLink to list fields (takes a field name and calls the previous method)

This is very useful for client side rendering templates, where you want to be able to jsLink your csr template to a field rather than to a web part, enabling the csr javascript to run on multiple views.


thomasberman:
Will make pull request for dev.

PR #435, opened 2016-03-03
Added SetJsLinkCustomizations site and list fields

thomasberman:
Added Field extension method to set jsLink
Added List extension to set jsLink on list field

This is very useful for client side rendering templates; being able to jsLink your csr template to a field rather than just a web part.


jansenbe:
Thx Thomas, appreciate the contribution!

PR #443, opened 2016-03-04
Adding Publishing Pages - Read only field fix

m-carter1:
Fix for checking if a field is not read only before trying to set the value when setting a files properties.

Not sure why this was previously commented out, but without this check we are unable to add a publishing page as it errors when trying to set a value on a read only field.


PaoloPia:
Hi Matt,
Thanks for your contribution. We forgot that flag. Thanks for the fix.

PR #445, opened 2016-03-07
Added Support for the following xml-elements: AllowedContentTypes/SharedFields/WelcomePageFields

martinlingstuyl:
Added Support for the following child elements of a Content Type inheriting from 'Document Set': AllowedContentTypes / SharedFields / WelcomePageFields

(Code Note: I added template.update(true) after every addition because if I didn't do that it would somehow only save the last item. Weird behavior!)


PaoloPia:
Hi Martin, 
Thanks for your contribution, we really appreciate it.

PR #466, opened 2016-03-14
Added FieldTitle tokenization for calculated fields

erwinvanhunen:


SuleymanA:
Hi Erwin,

I ran into this problem with calculated fields some time ago and reported it. Glad to see that there is a fix for it. I just generated a new template and I can see the tokenization for site field but not for list fields. Does this fix only apply to site fields? 

Thx.

jensotto:
This implementation introduces a major bug. If applying the provisioning template twice the calculated field will not work anymore. 
I guess this is because of some internal SharePoint manipulation when changing the SchemaXML. The provisioning template contains the field title (display name), but when provisioned to SharePoint, SharePoint translates it into the internal name and adds field references for each field. When applying the template for the second time it appears that SharePoint does not perform the translation of field title to field internal name. 
The formula looks like this after second deploy when I edit the field in the browser: =#NAVN?

Issue #470, opened 2016-03-15
The group description cannot be null or greater than 500 characters.


lanegoolsby:
This may actually be a bug in SP that was fixed in a CU. The environment I ran into this issue was quite behind on patches and I am fairly confident I covered this test case during development but can't say for certain. However, the fix is pretty benign. 

When provisioning a template, if a group description is not specified in the template it causes an error in SP that the group description cannot be null or over 500 characters. This only happens if the group already exists and the code is trying to modify the group instead of creating the group

PaoloPia:
Hi,
This one has been fixed by #471 and I will close it. Thanks for your contribution.

PR #471, opened 2016-03-15
Adding null/empty string check on group description before setting value

lanegoolsby:
I am not sure how the Documentation.md got included, I didn't change it and do not know where the changes came from. If you can't exclude it from the PR then I can reissue it so only the ObjectSecurity.cs is included. 

This PR addresses #470.

### DIFF
--- a/Core/Documentation.md
+++ b/Core/Documentation.md
@@ -1294,142 +1294,6 @@ Reorders content types on the list. The first one in the list is the default ite
 > **contentTypeNamesOrIds:** Content type names or ids to sort.
 
 
-#### SetLocalizationForContentType(Microsoft.SharePoint.Client.Web,System.String,System.String,System.String,System.String)
-Set localized labels for content type
-> ##### Parameters
-> **web:** Web to operate on
-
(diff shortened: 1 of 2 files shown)


jansenbe:
Thanks Lane, this is good addition that we'll merge. The reason you'll see the documentation.md file is because we've integrated the generation of this in the VS project file. Given that we have different configurations we'll need to revise that model. Thanks for making us aware of that.

As such I'll manually merge this PR but we'll ensure you'll get credited.

On a side note: we would like to make your company more visible in upcoming release notes but before we can include your company logo we would like to get your approval. Can you let us  know if you're OK with using the Threewill logo in future release notes? Thanks a lot!

lanegoolsby:
Thanks Bert. Feel free to use the ThreeWill logo. I am attaching the logo from our marketing team.

[ThreeWill-Logo-Inline.zip](https://github.com/OfficeDev/PnP-Sites-Core/files/179710/ThreeWill-Logo-Inline.zip)

jansenbe:
Thanks Lane. We've manually merged this

Issue #423, opened 2016-02-24
ObjectFeatures.cs excludes site scoped features on sub sites


lanegoolsby:
Title basically says it all, on line 24 there is an explicit check if the current context is a sub site or not. This causes issues when trying to provision a sub site that uses the Publishing features and the PE is not in charge of provisioning the root web for the site collection (e.g. the site collection already exists and need to provision a sub site). Is there a reason behind the exclusion?

PaoloPia:
Hi,
Thanks for your feedback, but I have to close this one because it is related to a closed PR.

PR #472, opened 2016-03-15
Removing check if site is a sub web or not.

lanegoolsby:
This fixes issue #423


jansenbe:
Hi Lane,

Thanks for your contribution. This is excellent code and definitely could do the thing you've described, however we can't accept this PR since it's going against one of our design principles that provisioning a site using a template can never change settings in other sites.

If you really need this you're definitely free to implement this functionality via an extension provider: this way you can still keep using the latest version of the engine while having your custom logic integrated.

Again thanks for your contribution, hope you understand the reasoning and we definitely hope to see more PR's from you in the future.

The PnP Core team.

Issue #202, opened 2015-11-18
Provisioning Engine does not create groups in sub sites


lanegoolsby:
In ObjectSiteSecurity.cs there is an explicit check for if the current site is a sub site or the root web and if its a sub site to not apply any of the settings defined in the <Security> section. Why is this? If a sub site isn't inheriting permissions then there shouldn't be an exclusion from creating groups for that site. 

        public override bool WillProvision(Web web, ProvisioningTemplate template)
        {
            if (!_willProvision.HasValue)
            {
                _willProvision = (template.Security.AdditionalAdministrators.Any() ||
                    template.Security.AdditionalMembers.Any() ||
                    template.Security.AdditionalOwners.Any() ||
                    template.Security.AdditionalVisitors.Any() ||
                    template.Security.SiteGroups.Any()) &&
                    !web.IsSubSite();
            }

            return _willProvision.Value;

        }

nickvdheuvel:
Will this be fixed? Or do I need to create custom code to make it possible to create groups in subsites (when permission inheritance is turned off)?

PaoloPia:
Hi,
the welcome page business logic is under reconstruction. Probably with the next monthly release it will be fixed. Thus, I will close this issue, waiting for the next master release. Thanks for your feedback and commitment.

lanegoolsby:
Hi @PaoloPia, this issue does not deal with welcome pages. This issue has to do with creating groups for sub sites that aren't inheriting permissions from the root web. Please reopen it. I just checked ObjectSiteSecurity.cs and the IF's in question still exist.

After speaking to some coworkers they are under the impression that the reason for the IF is because creating groups at a sub web is not a best practice but I don't quite understand the reasoning for that.

Regardless, if a site template includes groups but is being provisioned as a sub web then the groups need to be created, either at the web level or at the site level. Either adding a check to see if the web inherits permission from the parent site and creating them there if it does or automatically creating the groups at the site level and then setting the site to use those permissions. 

I am actually working having to implement some of this anyways, so if you can provide guidance on if creating the groups at the sub web level is acceptable or if groups should only be created at the root web then I would be happy to add my changes and issue a PR.

PaoloPia:
Oops  ... I guess I wrote the comment in the wrong issue :) ... I'm sorry. Thanks for letting me know. We will come back with some feedbacks and guidance, as soon as possible.

lanegoolsby:
[...]

PR #473, opened 2016-03-15
Adding logic to create all groups for sub web at the root web level i…

lanegoolsby:
…nstead.

This resolves the core of Issue #202. I will create a new issue to track that the engine doesn't allow admins/developers a way to specify if a site should or should inherit security from the parent.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
@@ -14,17 +14,22 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
         {
             get { return "Site Security"; }
         }
-        public override TokenParser ProvisionObjects(Web web, ProvisioningTemplate template, TokenParser parser, ProvisioningTemplateApplyingInformation applyingInformation)
+        public override TokenParser ProvisionObjects(Web Web, ProvisioningTemplate template, TokenParser parser, ProvisioningTemplateApplyingInformation applyingInformation)
         {
             using (var scope = new PnPMonitoredScope(this.Name))
             {
-
+                Web web;
+                bool usingRootWeb = false;
 
                 // if this is a sub site then we're not provisioning security as by default security is inherited from the root site
-                if (web.IsSubSite())
+                if (Web.IsSubSite())
                 {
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi Lane,

Thanks for your contribution. This is excellent code and definitely could do the thing you've described, however we can't accept this PR since it's going against one of our design principles that provisioning a site using a template can never change settings in other sites.

If you really need this you're definitely free to implement this functionality via an extension provider: this way you can still keep using the latest version of the engine while having your custom logic integrated.

Again thanks for your contribution, hope you understand the reasoning and we definitely hope to see more PR's from you in the future.

The PnP Core team.

hyankov:
Hi,

Well, it's not really "other sites", it's the parent site of the subweb. Because, you know, that's how SharePoint works when you provision groups - they go to the site collection level.

Let's look at the following template, being executed on a **subsite with broken inheritance**:

```
<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2015/12/ProvisioningSchema">
  <pnp:Preferences Generator="OfficeDevPnP.Core, Version=2.3.1604.0, Culture=neutral, PublicKeyToken=3751622786b357c2" />
  <pnp:Templates ID="CONTAINER-TEMPLATE-XXX">
    <pnp:ProvisioningTemplate ID="TEMPLATE-XXX" Version="1">
      <pnp:Security>
        <pnp:SiteGroups>
          <pnp:SiteGroup Title="My Group" Owner="SHAREPOINT\system" AllowMembersEditMembership="false" AllowRequestToJoinLeave="false" AutoAcceptRequestToJoinLeave="false" OnlyAllowMembersViewMembership="true" />
        </pnp:SiteGroups>
        <pnp:Permissions>
          <pnp:RoleAssignments>
            <pnp:RoleAssignment Principal="My Group" RoleDefinition="Full Control" />
          </pnp:RoleAssignments>
        </pnp:Permissions>
      </pnp:Security>
    </pnp:ProvisioningTemplate>
  </pnp:Templates>
</pnp:Provisioning>
```

It should create a 'My Group' group on the Site Collection level and then give 'My Group' 'Full Control' permissions **on the subweb, yes**? What is the impact on the Site Collection - a new container being added, without any permission change done to the Site Collection root web, because the subsite inheritance is broken.

If the subsite permissions were not broken, then I would agree with your rule, however, in this specific case I don't think your concept aligns with the way SharePoint works!

So, how should it work (in my opinion): you should be checking if the subsite has broken inheritance, if it does, you should NOT ignore the Security element of the PnP template schema, because it does not really affect the parent site.

Thanks for listening.

PR #480, opened 2016-03-18
Added Order for Webpart provisioning

m-carter1:
Although the template contained the order for the webparts it was not coming through when provisioning this was because it was not being set, so it has now been added


jansenbe:
Thx for this update Matt :-)

PR #481, opened 2016-03-18
Fix for List already exists bug when the list title is already in use

m-carter1:
When provisioning it was only check for an existing list by the url but if a list has the same title but different url it would error so it now checks for the list by url and title


PaoloPia:
Thanks Matt. We really appreciate your efforts!

PR #483, opened 2016-03-21
Added DeleteContentTypeByName

thomasberman:
Added method to delete a content type from the web by name. Will add byId later.


jansenbe:
Thanks Thomas. We've processed your PR.

Issue #484, opened 2016-03-21
Provisioning Engine - Publish file as major version


luismanez:
As far as I've experienced, there's no way to publish a File in Major version from the Provisioning engine. Perhaps depends on how the Lib is configured, but for instance, with the OOB Master page gallery, the files (Layouts, Master pages, Display templates) are provisioned but not published as major versions. This is the pnp:File node I'm using:

```xml
<pnp:File Src="..\Assets\DisplayTemplates\Content Web Parts\Custom_Item_News.html" Folder="~masterpagecatalog/Display Templates/Content Web Parts" Overwrite="true">
          <pnp:Properties>
            <pnp:Property Key="Title"  Value="Custom Default Item" />
            <pnp:Property Key="ContentTypeId" Value="{contenttypeid:Item Display Template}" />
            <pnp:Property Key="MasterPageDescription"  Value="Displays the Search Results default item in a custom way." />
            <pnp:Property Key="UIVersion"  Value="15" />
            <pnp:Property Key="TemplateHidden"  Value="0" />
            <pnp:Property Key="TargetControlType"  Value=";#Content Web Parts;#" />
            <pnp:Property Key="DisplayTemplateLevel"  Value="Item" />
            <pnp:Property Key="ManagedPropertyMapping"  Value="'Link URL'{Link URL}:'Path','Line 1'{Line 1}:'Title','Line 2'{Line 2}:'','FileExtension','SecondaryFileExtension'" />
          </pnp:Properties>
        </pnp:File>
```

I'm gonna do a Pull Request with a possible temporary solution to this, but IMHO the right way to fix it should be to allow a **Level** attribute in the pnp:File xml node.

MarkStokes-outlook:
I would also like the ability to provision my master laws and page layouts as Published and Approved.

MarkStokes-outlook:
*master pages

luismanez:
with latest PnP Schema, now is possible to set the Level of the File:
https://github.com/OfficeDev/PnP-Provisioning-Schema/blob/master/ProvisioningSchema-2016-05.md

so we can close this issue.
/cc @VesaJuvonen @PaoloPia

bartdonninger:
Hello,

The latest schema does indeed offer the attribute Level on File and Directory elements. But when i want to provision my masterpages to the MasterPage Gallery they do not get published to a 1.0 version.

Can anyone confirm if this is still an issue for them as well?

sandermoes:
[...]

PR #485, opened 2016-03-21
This is a propossed solution for issue #484

luismanez:
That way you can add a File property node with name "_Level" and a string value like the FileLevel enum (Published, Draft, Checkout)
I guess the right way should be to add a new attribute to the pnp:File node, but that requires to change the PnP Schema.

Note: not sure why GitHub is detecting so many changes in the file. Just added: 190-191, 222-227, 285-289. Rest of the file is not touched.


PaoloPia:
Thanks, I've added the Level attribute to the File element within the new and upcoming schema 2016-05. See here: https://github.com/OfficeDev/PnP-Provisioning-Schema/pull/93. I will close this PR, because we will support this capability officially. But thanks again for sharing and for contributing.

luismanez:
That's great! Thanks to you Paolo.

PR #487, opened 2016-03-22
Fix for #323 Workflows not provisioning correctly

m-carter1:
We were experiencing this issue again for another site where the workflow subscriptions weren't correctly provisioned, now fixed by copying CreatedBySPD property definition value when creating the subscription


PaoloPia:
Hi Matt,
Thanks for your contribution, the PR has been merged.

PR #492, opened 2016-03-24
Cater for MultiChoice field.

warrenroscoe:
Added support for setting properties of multichoice field when adding files.
New to git, so ... please be patient if I got anything wrong. Should be one file change only (objectfiles.cs).


PaoloPia:
Hi, Thanks for your contribution. We really appreciate your efforts. I will not merge this PR because it changes also the project version, but I will merge the code change and you will be credited for your contribution. Thanks.

PR #505, opened 2016-04-06
Adding override method to allow control over OnlyAllowMembersViewMemb…

lanegoolsby:
…ership


jansenbe:
Thanks Lane, I'll have a look at this one.

jansenbe:
Hi Lane, 
I manually processed this PR as I could not take over your code since you changed the method signature which will result in breaking folks who where using that method. I added an additional optional parameter to the method signature which defaults to false.

lanegoolsby:
I left the original method signature in place and just called the override with the new param set to false. But your approach works, too. Tomato, Tomato. Thanks Bert!

PR #527, opened 2016-04-26
Implemented an array of writeable readonly-fields to replace the check in SetFileProperties

olemp:
Makes it easier and more readable if we need to extend with more properties.


PaoloPia:
Thanks Ole Martin, I agree with the proposed refactoring. Thanks for your contribution.

PR #536, opened 2016-05-05
Added catch and retry for connection error when saving new Workflow Definitons

m-carter1:
This pull request adds a retry when saving a new workflow definition as this is the area we are seeing the random "The underlying connection was closed: A connection that was expected to be kept alive was closed by the server." error appear.

We did look at extending ExecuteQueryRetry to catch the error but the WorkflowDefinition object needs to be re-created when retrying else it will fail to save correctly.

Also added an ExecuteQueryRetry after publishing the definition as it was missing and is required.


PaoloPia:
Thanks Matt,
We really appreciate your efforts! Well done!

PR #558, opened 2016-05-17
Ensure folder fix1

biogenez:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes

Bug desciption:
For example, we want to ensure folder path "AAAAAAAAA/BBB/C".
The website ("https://myportal.com/sites/test") contains two libraries: "AAA" and "AAAAAAAAA".
The root folder server relative path:
 for library "AAA": "/sites/test/AAA"
 for library "AAAAAAAAA": "/sites/test/AAAAAAAAA"
 
The variable folderServerRelativeUrl = "/sites/test/AAAAAAAAA/BBB/C" 

So, If we call the extension method web.EnsureFolderPath("AAAAAAAAA/BBB/C"), 
the folder may be created in the wrong library (with name "AAA"), because the string "/sites/test/AAAAAAAAA/BBB/C" starts with "/sites/test/AAA".
(as the result we have the incorrect folder server relative path: "/sites/test/AAA/AAAAAA/BBB/C")

This PR fix this bug. As a result, the path ensured properly: (https://myportal.com/sites/test/AAAAAAAAA/BBB/C)

### DIFF
--- a/Core/OfficeDevPnP.Core/AppModelExtensions/FileFolderExtensions.cs
+++ b/Core/OfficeDevPnP.Core/AppModelExtensions/FileFolderExtensions.cs
@@ -382,10 +382,9 @@ namespace Microsoft.SharePoint.Client
             web.Context.ExecuteQueryRetry();
 
             List containingList = null;
-            var searchFldUrl = folderServerRelativeUrl.EndsWith("/") ? folderServerRelativeUrl : folderServerRelativeUrl + "/";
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi Alexander, 

Thanks for this bug fix. Looking at the code it makes sense, but we'll need to first test it before it can be merged. We'll try to get it merged before the June release.

PR #560, opened 2016-05-18
Localization export performance

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no

Exporting localization resources is very slow. To improve performance this change will skip resource export of Fields, Content Types and Lists if they exist in the BaseTemplate

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectContentType.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectContentType.cs
@@ -404,13 +404,21 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                     if (creationInfo.PersistMultiLanguageResources)
                     {
 #if !SP2013
-                        if (UserResourceExtensions.PersistResourceValue(ct.NameResource, string.Format("ContentType_{0}_Title", ct.Name.Replace(" ", "_")), template, creationInfo))
+                        if (creationInfo.BaseTemplate != null)
                         {
-                            newCT.Name = string.Format("{{res:ContentType_{0}_Title}}", ct.Name.Replace(" ", "_"));
-                        }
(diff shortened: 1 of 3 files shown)


jansenbe:
Hi Michael,

Thanks for this good change. We'll evaluate and test it in the coming weeks and try to get it merged in before the June release.

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectContentType.cs
```diff
@@ -404,13 +404,21 @@ private IEnumerable<ContentType> GetEntities(Web web, PnPMonitoredScope scope, P
...
 #if !SP2013
-                        if (UserResourceExtensions.PersistResourceValue(ct.NameResource, string.Format("ContentType_{0}_Title", ct.Name.Replace(" ", "_")), template, creationInfo))
+                        if (creationInfo.BaseTemplate != null)
                         {
-                            newCT.Name = string.Format("{{res:ContentType_{0}_Title}}", ct.Name.Replace(" ", "_"));
-                        }
-                        if (UserResourceExtensions.PersistResourceValue(ct.DescriptionResource, string.Format("ContentType_{0}_Description", ct.Name.Replace(" ", "_")), template, creationInfo))
-                        {
-                            newCT.Description = string.Format("{{res:ContentType_{0}_Description}}", ct.Name.Replace(" ", "_"));
+                            int index = creationInfo.BaseTemplate.ContentTypes.FindIndex(c => c.Id.Equals(ct.StringId));
```
erwinvanhunen: This would exclude situations where a base template contains a CT with a certain id and a name and the template contains a different title for the CT but it has the same ID

jansenbe:
Hi Michael, I accidently implemented similar logic for fields a while ago...the implementation is slightly different and I've now also done that for CT's...meaning we're not merging this PR but will credit you for the change as your change was the first :-)

PR #561, opened 2016-05-18
IsSubSite not working when using admin context

thlbrg:
| Q               | A
| --------------- | ---
| Bug fix?        | Yes
| New feature?    | No
| New sample?      | No
| Related issues?  | No

Fix for IsSubSite, previously when using the -admin context, and then opening another site collection, the new site collection would be compared with the -admin root web id when checking to see if it's a sub site, and always returning true. 

Reproduce: 
string tenantName = "mytenant";

using (ClientContext ctx = new ClientContext($"https://{tenantName}-admin.sharepoint.com"))
{
  ctx.Credentials = credentials;
  var tenant = new Tenant(ctx);
  Site site = tenant.GetSiteByUrl($"https://{tenantName}.sharepoint.com");
  Console.WriteLine(site.RootWeb.IsSubSite());
}

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,42 @@
+Thank you for reporting an issue or suggesting an enhancement. We appreciate your feedback - to help the team to understand your needs, please complete the below template to ensure we have the necessary details to assist you. If you have a actual question, we would ask you to use PnP Yammer group at http://aka.ms/OfficeDevPnPYammer. Thanks!
+
+#### Which PnP repository to report the issue?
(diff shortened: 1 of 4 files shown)


jansenbe:
Hi thlbrg,

Looks like a good change but we'll need to first test it before we can merge it. Thx for your contribution.

jansenbe:
Hi thlbrg,

Testing showed that this implementation did not work. Can you please cross check and verify this...maybe we've missed something but our IsSubWebTest() test case fails with your implementation.

thlbrg:
Interesting, i'm getting all green here, will take a look when I've more time. 
2016-06-09 13_07_47-testrun

PR #562, opened 2016-05-18
Sitegroups

rjbooden:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no 
| New sample?      | no
| Related issues?  | have not checked if there are related issues

fix: check if a group exists with a unparsed title is inconsistent with adding the group with a parsed title. This can result in an exception when a group including {sitename} is used.

changed: excessive parsing to parsing once

Fix: if the principal of a role assignment is a sharepoint group, the name of the principal should also replace the site title as is done for sitegroups. If not an inconsistency can occur causing an exception when provisioning the template, the group/principal might not exist.

Fix: don't add a role assignment for a principal if the principal is a sharepoint group when sharepoint groups are not included. Provisioning will fail if the group/principal does not exist, and they will probably not exist (other than the default groups) if the are not in the provisioning template.

changed: only call ReplaceGroupTokens for sharepoint groups, otherwise there is no need for this method. This also prevents issues in rare cases where the loginName includes the site title or one of the default group names.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
@@ -5,6 +5,7 @@ using Microsoft.SharePoint.Client;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
 using User = OfficeDevPnP.Core.Framework.Provisioning.Model.User;
 using OfficeDevPnP.Core.Diagnostics;
+using Microsoft.SharePoint.Client.Utilities;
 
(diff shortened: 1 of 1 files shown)


rjbooden:
I was thinking about the difference between the Groups and SiteGroups in the web. As I now skip the role assignments for SharePoint groups when SiteGroups are not included, this could be an issue for a sub site, with broken role inheritance where groups have different role assignments than in the parent web.

I guess skipping the role assignments when SiteGroups are not included should only be done if the current web is the rootweb. Or it should never be skipped, and it is up to the developers to always include the SiteGroups when provisioning the rootweb.

rjbooden:
Will close the pull request and remove the skipping of role assignments when sitegroups are not included. Developers should include the sitegroups for the rootweb if there are sharepoint groups other than the default groups are added.

PR #564, opened 2016-05-20
TaxonomyField tokenization fix

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues? | [#572](https://github.com/OfficeDev/PnP-Sites-Core/issues/572)

This PR fixes three issues.

1. The termsetid token wasn't set in the field schemaxml because an ExecuteQuery was missing after store.GetTermSet()
2. Only site columns were tokenized. Now also list column taxanomy fields are tokenized.
3. If the TermGroup is a Site Collection Term Group, the group name was inserted in the TermSetId token, and caused the IsFieldXmlValid to fail when provisioning. This is changed to insert the "{sitecollectiontermgroupname}" token.


jansenbe:
Thx Michael. We'll evaluate and test this change.

PaoloPia:
Thanks Michael,
We really appreciate your efforts. The PR has been merged.

PR #565, opened 2016-05-20
Sitegroups

rjbooden:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | have not checked if there are related issues

fix: check if a group exists with a unparsed title is inconsistent with adding the group with a parsed title. This can result in an exception when a group including {sitename} is used.

changed: 'excessive' parsing to parsing once

fix: if the principal is a sharepoint group, the name of the principal should also replace the site title as is done for sitegroups. If not an inconsistency can occur causing an exception when provisioning the template, the group/principal might not exist.

fix: don't add a role assignment for a principal if the principal is a sharepoint group when sharepoint groups are not included. Provisioning will fail if the group/principal does not exist, and they will probably not exist (other than the default groups) if the are not in the provisioning template.

changed: only call ReplaceGroupTokens for sharepoint groups, otherwise there is no need for this method. This also prevents issues in rare cases where the loginName includes the site title or one of the default group names.

fix: when a template is generated from a site that has a subsite with unique permissions, when provisioning an exception can occur on a non existing group owner. Also groups not used in the site, should not be included in the template.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
@@ -5,6 +5,7 @@ using Microsoft.SharePoint.Client;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
 using User = OfficeDevPnP.Core.Framework.Provisioning.Model.User;
 using OfficeDevPnP.Core.Diagnostics;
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you very much, Ronald!

After long deliberation and testing, we decided to partially merge you PR. We think that even if a sitegroup has no role assignment, maybe someone still needs to move the group between sites during extraction and provisioning. That's why we decided to not include that commit in this PR.

PR #592, opened 2016-06-02
Updated NuGet dependencies to use on-prem versions and not SharePoint…

robi26:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no 

The SharePointPnPCore2013 and SharePointPnPCore2016 NuGet packages for on-prem reference the AppForSharePointOnlineWebToolkit which is intended for use in SharePoint online add-ins. Changed dependency to AppForSharePointWebToolkit which is intended for on-prem.

Generally it would help a lot to get rid of the AppForSharePointWebToolkit dependency. The PnP project does not really depend on it and if people want to use the SPContext and TokenHelper from the AppForSharePointWebToolkit package they can add the nuget package themself. Please (re-)consider this option. Thanks.

### DIFF
--- a/Core/SharePointPnPCore2013.2.4.1605.0.nuspec
+++ b/Core/SharePointPnPCore2013.2.4.1605.0.nuspec
@@ -17,7 +17,7 @@
         <dependencies>
             <dependency id="Microsoft.AspNet.WebApi.Client" version="5.2.3" />
             <dependency id="Microsoft.AspNet.WebApi.Core" version="5.2.3" />
-            <dependency id="AppForSharePointOnlineWebToolkit" version="3.1.2" />
+            <dependency id="AppForSharePointWebToolkit" version="3.1.2" />
         </dependencies>
         <frameworkAssemblies>
             <frameworkAssembly assemblyName="System" targetFramework="" />
--- a/Core/SharePointPnPCore2016.2.4.1605.0.nuspec
+++ b/Core/SharePointPnPCore2016.2.4.1605.0.nuspec
@@ -17,7 +17,7 @@
         <dependencies>
             <dependency id="Microsoft.AspNet.WebApi.Client" version="5.2.3" />
             <dependency id="Microsoft.AspNet.WebApi.Core" version="5.2.3" />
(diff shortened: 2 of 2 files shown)


jansenbe:
Hi Stephan,

Thanks for this PR...your reasoning indeed is logical...but we've taken a slightly different (also logical I hope :) ) approach. All our NuGet packages reference the AppForSharePointOnlineWebToolkit NuGet package because that one is only bringing the SharePointContext and TokenHelper classes and not assembly references. We've included the assembly references for 2013 and 2016 in our PnP nuget packages.

Have you seen problems with our approach?

I'll close this PR for now, but feel free to open a new issue if you see this as a problem. Updating the nuspec files will not help, they're more there for reference.

pascalberger:
For our approach this is still highly confusing for the developers. We have wrapped SharePointContext and TokenHelper in our own assembly which we include in our projects. Therefore our own projects don't need any SharePointContext / TokenHelper, just the references. Now considering in a SharePoint 2013 Add-In, with PnP having a dependency to AppForSharePointOnlineWebToolkit, this adds SharePointContext and TokenHelper classes back to the project, but they are neither used by PnP nor should they be used in our application... On the other hand we still need to add the references, which are required by PnP and our application, manually to the project. IMHO it doesn't make any sense to add an artificial dependency to something which is not required at all, but instead don't add dependencies to required references. Cleanest solution would be to introduce a NuGet package containing the SharePoint references, and only the SharePoint references, and add this as a dependency, because this is technically the only existing dependency. Until then I would prefer to have no dependency at all to allow developers to architecture the Add-In the way how it fits for their purpose.

jansenbe:
Hi Pascal,

I understand this might be confusing...but on the other hand it's also confusing to not have these files included. For SP projects the files are added as part of the SP project but for console applications (like Azure web jobs) they're not...in the current model you'll always get the files when you use a PnP package. We reasoned that dropping those 2 files is easier then grabbing them when you need them.

In the end both models can work....but we need to make a choice.

robi26:
I second @pascalberger arguments. 
BTW, everytime I upgrade the PnP package the files are re-added.

PR #597, opened 2016-06-03
Allow adding navigation nodes at beginning or end

casselc:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no

Adds `asLastNode` parameter to the `AddNavigationNode` method to control whether new nodes are added to the beginning or end of the collection.`asLastNode` defaults to `true` to match the current behavior.


jansenbe:
Hi casselc, this is a useful change and we'll get it merged.

Would it be possible to update your GitHub profile so that we can give you credits using your real name instead of @casselc?

PR #603, opened 2016-06-06
Add ListExists() based on List's url path

rroman81:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no

Added a few ListExists() test cases, and wrote an additional overload that determines if the List exists based on url path.

This is handy to have as when using CreateList method overload you can hard-code url path and at times when lists are renamed, it's hard to determine by title if list already exists.


jansenbe:
Thx Roman, looks good. We'll do some testing and then get it merged.

jansenbe:
Thx Roman, was all good. I'll merge these changes and apologies for the late processing...I'm blaming summer vacation for that :-)

PR #632, opened 2016-06-17
Fix: FieldRefs containing replacement tokens are ordered wrong

okms:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

Fixing issue where we have fields defined with internal names containing tokens and re-ordering of these. The bug manifested itself through OOB fields getting sorted before our custom fields for no apparent reason. The unittest reflects this scenario.

example 
This field 
`<Field ID="{dd6b7dae-1281-458d-a66c-01b0c7b7930b}" Name="{parameter:TestFieldPrefix}TestField" DisplayName="TestField" Type="Note" Group="PnP" Description="" />
`

Would end up as the last field when referenced like this

```
<pnp:ContentType ID="0x010100503B9E20E5455344BFAC2292DC6FE805" Name="Test Content Type" Description="" Group="PnP}">
    <pnp:FieldRefs>
        <pnp:FieldRef ID="dd6b7dae-1281-458d-a66c-01b0c7b7930b" Name="{parameter:TestFieldPrefix}TestField" />
        <pnp:FieldRef ID="53101f38-dd2e-458c-b245-0c236cc13d1a" Name="AssignedTo" />
    </pnp:FieldRefs>
</pnp:ContentType>
```

A work around could of course be to not use tokens when adding fieldrefs, but that defeats the purpose of supporting tokens in the first place


jansenbe:
Thx Ole Kristian,

jansenbe:
Nice fix Ole. Implemented...now that vacation is over :-)

PR #638, opened 2016-06-22
Provided List.GetWebRelativeUrl() which gets the web relative URL of a list.

sebastianrogers:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | none

A utility method which allows you to get the 'Web Relative URL' of a list suitable for use in the existing Web.GetListByUrl() function.

When writing code to export a list definition to a file, i.e. XML or JSON, this value is useful as it allows the file to act as an import file.

The Web Relative URL of a list is actually a unique identifier for the list as it can't be changed once created, unlike a title, which makes it extremely useful for examining SharePoint web's to determine if the list already exists.


jansenbe:
Hi Sebastian,

jansenbe:
Thx for this change Sebastien, we'll merge it.

PR #639, opened 2016-06-22
Added URL decoding in the Tokenize method. If an object has been crea…

estruyf:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes

Added a small fix to the Tokenize method. If a URL contains spaces, SharePoint sometimes returns them as encoded strings with "%20" instead of real spaces. This results in a URL without a token. With this fix, the URL will get decoded first.


erwinvanhunen:
Thanks Elio!

PR #640, opened 2016-06-23
Master page URL token fix when provisioning template to a sub-site.

estruyf:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

This fixes an issue with replacing the {masterpagecatalog} with the correct master page catalog URL. 

At the moment the token gets replaced with the master page catalog URL of the web you are provisioning the template at. If the web is a sub-site. This returns the path to the master page gallery URL of that web and causes a file not found issue (because the reference is set to /site/sub-site/_catalogs/masterpage/custom.master).

With this fix, it will be retrieved from the rootweb if the current web is a sub-site. The returned URL is: _catalogs/masterpage.


jansenbe:
Thx Elio,

PaoloPia:
Hi Elio,
Thanks for your contribution. The change looks good. Unfortunately, because of a conflict, I had to merge it manually and this PR will be closed. Nevertheless, you will be credited in the monthly release. Thank you and see you soon :-) ...

PR #651, opened 2016-06-29
Move field ref code inside null check

eric-scott-owens:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes null check not covering all uses of the "field" variable

Improved null check which did not cover all uses of the nullable field variable when creating list instances.


Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
```diff
@@ -123,17 +123,17 @@ public override TokenParser ProvisionObjects(Web web, ProvisioningTemplate templ
                             foreach (var fieldRef in listInfo.TemplateList.FieldRefs)
                             {
                                 var field = rootWeb.GetFieldById<Field>(fieldRef.Id);
-                                if (field != null)
+                                if (field == null) continue;
```
s-KaiNet: Should we continue when field is not found? What if you copied inaccurate id by mistake or content type hub doesn't work as expected. You will never know that some of your `fieldrefs` are not updated (because they are wrong).

eric-scott-owens: True. I'm in favor of logging it and still letting it complete the process. Then you know something wasn't quite right but the rest of the work completes. What do you think?

s-KaiNet: Agreed. Continue and logging is quite good choice.

jansenbe:
Hi Eric,

This change makes sense and we'll look into this later on. Given holidays this most likely will not be processed before the next master merge.

jansenbe:
Merged. Thx for your patience as it took a while.

PR #699, opened 2016-08-10
Possible solution to Issue #698 (Custom ProvisiningExtensibilityHandl…

### LINKED ISSUE #698: Custom ProvisiningExtensibilityHandler GetTokens method is never invoked.
I've created a custom ProvisiningExtensibilityHandler implementing the IProvisioningExtensibilityHandler interface. The provider is registered OK, and the Extract and Provision methods are invoked as expected, however, the GetTokens method is never invoked. I think this is because some code in the ExtensibilityManager is still expecting the old interface "IProvisioningExtensibilityProvider"

I think I solved the issue in my local, and I'm gonna do a Pull Request with my propossed solution.

#### Category
[X] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ ] SharePoint 2013

Custom ProvisiningExtensibilityHandler GetTokens method is invoked.

#### Observed Behavior
Custom ProvisiningExtensibilityHandler GetTokens method is NOT invoked. If you set a breakpoint to that method, you'll see is not invoked never. Seems is because some code in the ExtensibilityManager is expecting an object with the old interface "IProvisioningExtensibilityProvider".

#### Steps to Reproduce
_If you are reporting a bug please describe the steps to reproduce the bug in sufficient detail to allow testing. Only way to fix things properly, is to have sufficient details to reproduce it. If you are making a suggestion, you can delete this section._

#### Submission Guidelines
_Delete this section after reading_
- All suggestions or bugs are welcome, please let us know what's on your mind.
- If you are reporting any issues around PnP Core Component, please reproduce the issue with latest release.
- If you are reporting issue around PnP Provisioning Engine, please share the xml template, if possible.
- If you are reporting an issue around any of the samples, please ensure that you have clear reference on the sample and possibly code file, which should be fixed.
- If you have a general question, please use the [PnP Yammer Group](http://aka.ms/OfficeDevPnPYammer).
- Remember to include sufficient details and context.
- If you have multiple suggestions or bugs please submit them in seperate bugs so we can track resolution.

Thanks for your contribution! Sharing is caring.

luismanez:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| Related issues?  | fixes #698 

This PR is a propossed solution for issue #698 

This code was wrong, as doing the Union with OfType<Provider> resulted in an empty collection, as the custom handler was implmenting the new interface.

```c#
foreach (var handler in template.ExtensibilityHandlers.Union(template.Providers).OfType<Provider>()
```

Also, in the *ExtensibilityManager* class, there was an issue with this header:

```c#
public IEnumerable<TokenDefinition> ExecuteTokenProviderCallOut(ClientContext ctx, Provider provider, ProvisioningTemplate template)
```
as the *Provider* object was wrong, and it has to be a *ExtensibilityHandler* object.

I'm not checkeing if the custom Handler is using the new or the old interface here, as the Tokens thing is only supported by the new interface, so here we never hit with an old Provider object.

Hope that makes sense.

Thanks!

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/Extensibility/ExtensibilityManager.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/Extensibility/ExtensibilityManager.cs
@@ -25,7 +25,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.Extensibility
         /// <exception cref="ExtensiblityPipelineException"></exception>
         /// <exception cref="ArgumentException">Provider.Assembly or Provider.Type is NullOrWhiteSpace></exception>
         /// <exception cref="ArgumentNullException">ClientContext is Null></exception>
-        public IEnumerable<TokenDefinition> ExecuteTokenProviderCallOut(ClientContext ctx, Provider provider, ProvisioningTemplate template)
+        public IEnumerable<TokenDefinition> ExecuteTokenProviderCallOut(ClientContext ctx, ExtensibilityHandler provider, ProvisioningTemplate template)
         {
             var _loggingSource = "OfficeDevPnP.Core.Framework.Provisioning.Extensibility.ExtensibilityManager.ExecuteTokenProviderCallOut";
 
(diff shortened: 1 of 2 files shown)


PaoloPia:
Hi Luis,
Thanks for your contribution. The proposed solutions makes sense and looks good. Merged!

luismanez:
Many thanks Paolo !

PR #701, opened 2016-08-12
use isDirty for CommandUIExtension

dbeaulieu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | n/a

CommandUIExtension updates should be treated like updating any other property.  This was causing the update to get sent in the next clientContext batch.


jansenbe:
Hi Dan,

Nice catch :-)  Thanks for fixing this.

PR #709, opened 2016-08-16
Security and Site Policy validations

srinutamada:
Hi Bert,

We have added security and site policy validator to the PnP site core solution.

Thanks and Regards,
Srinu Tamada


srinutamada:
Hi Bert,

We have implemented pages, files and workflow validators to the PnP site core and added to the same pull request. Could you please review and branch?

Regards,
Srinu Tamada

PR #728, opened 2016-09-05
Openxml fixes

ivanvagunin:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| New sample? | no |
| Related issues? |  |

Fixes for incorrect folders handling in OpenXMLConnector. Currently there are two issues with that: 1) mapped folders are not read correctly when opening package 2) mixing forward and backward slashes in file path when adding files to package causes file duplication under the same mapped path.

Also added 2 unit tests to validate those fixes (OpenXMLGetFilesFromFolder, OpenXMLFileDuplicationTest) (did not modify other tests)


PaoloPia:
Thanks Ivan,
I will look into it and I'll let you know.

ivanvagunin:
Thanks, Paolo! Let me know if you need any help, but I could reproduce those quite easily.

PaoloPia:
Hi Ivan,
I looked into this PR. One question that I have about the fix for issue 2) is: have you tested it against an old .PNP file (which has subfolders and files with different / \ values in the folder) in order to check for backward compatibility? Thanks.

ivanvagunin:
Hi Paolo,

I did not test much for backward compatibility (only one unit test that was already implemented). Is it supposed to be 100% backward compatible? One thing I've noticed about current implementation that it was crashing in a scenario when you try to add new files to "old" package, since file map is not initialized. 
I can implement more unit tests for backward compatibility, just need more on how it is supposed to work. E.g. when you add files to old package is it supposed to overwrite all files (include all files in the map and create flat structure under Files folder) or only new files should be added to map?

Br, Ivan

ivanvagunin:
Hi Paolo,
I've added one compatibility test (OpenXMLLoadTemplateOriginal2) for Open XML and it is failing. 
The reason is that old Package is opened for Read access, but PnPPackage.EnsurePackagePartWithRelationshipType tries to modify it, since it does not have files map part. Not sure what should be a proper fix for that. Could you please check?

PaoloPia:
Ivan,
I'm digging into this PR and into the issue about the missing FilesMap ... I made a fix to the issue that you highlighted and I merged all into DEV. Please double-check my changes in your scenarios and let me know if now they work. Thanks.

PR #749, opened 2016-09-10
Fix to folder creation issue on lists with no Title field available

erwinvanhunen:


jansenbe:
Looks good Erwin, I'll merge it already and kick off some unit tests

PR #787, opened 2016-10-06
Updated to allow for the pageName to include relative path info.

mrpullen:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes? |
| New sample? | no |
| Related issues? | fixes #X, partially #Y, mentioned in #Z |

Please describe the changes in this PR. Sample description or details around bugs which are being fixed.

When using OfficePnp Powershell and adding publishing pages (Add‑SPOPublishingPage), it is much easier to allow the relative path to be supplied as part of the page name parameter. The name parameter "folder/page.aspx" should provision a page in /Pages/folder/page.aspx path. Currently since the / is an invalid character we instead get /Pages/folder-page.aspx. Also since the / character is a standard separator seems like an intuitive addition.

### DIFF
--- a/Core/OfficeDevPnP.Core/AppModelExtensions/PageExtensions.cs
+++ b/Core/OfficeDevPnP.Core/AppModelExtensions/PageExtensions.cs
@@ -1142,7 +1142,18 @@ namespace Microsoft.SharePoint.Client
             {
                 title = pageName;
             }
-
+            //Allow users to specify the filename as a relativepath within the pages library ie: folder/page.aspx would provision
+            //a new publishing page within /Pages/folder/ named page.aspx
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi Michael,

Thanks for this PR. Since the AddPublishingPage in PnP sites core already allows for a folder object to be passed in I'm not going to merge this PR. In my view it's better to handle this inside the Add‑SPOPublishingPage implementation, looping in @erwinvanhunen for his comments as well.

For now I'm gonna close this PR, but depending on the feedback from Erwin we can re-open this PR later on.

Thanks again for your contribution, hope the above explanation makes sense.

Issue #744, opened 2016-09-08
Calculated fields with non existing field references failing to provision


Pieter-Veenstra:
#### Which PnP repository to report the issue?
- PnP Sites Core - https://github.com/OfficeDev/PnP-Sites-Core
#### Category

[X] Bug
[ ] Enhancement
#### Environment

[X] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013
#### Expected or Desired Behavior

When a calculated field references a field in the Formula that hasn't been created yet (because it is created lower in the template XML) then the creation of the calculated field should be delayed.
#### Observed Behavior

When a calculated field references a field in the Formula that hasn't been created yet (because it is created lower in the template XML) then the created of the calculated field fails with the following error:

Apply-SPOProvisioningTemplate : The field was found invalid:  {fieldtitle:Done_x0020_Date}
#### Steps to Reproduce

used latest release available. Exported a site with calculated field and imported it to a clean site. Reproducing this issues might need a bit of luck as there isn't a specified order in the fields being exported.

jansenbe:
Fixed by #817

PR #802, opened 2016-10-14
Fix parsing of calculated fields with references to fields not yet created

### LINKED ISSUE #537: Applying template - field order with dependent fields
I created a template from an existing site and when applying to another site an exception was thrown when trying to create a calculated column because referenced fields had not yet been created.

> Create site with calculated column referencing other fields
> Get template from site
> Apply template (error may be thrown if the calculated field is created before the referenced fields)

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #537 and #744 

Please see #537 and #744 for a description of the problem.

Instead of adding a FieldTitleToken to the TokenParser after a field is created, this PR adds the token in the constructor of the TokenParser. This will make the tokens available when parsing a calculated field where the referenced field(s) are not yet created.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectField.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectField.cs
@@ -203,9 +203,6 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 web.Context.Load(field, f => f.TypeAsString, f => f.DefaultValue, f => f.InternalName, f => f.Title);
                 web.Context.ExecuteQueryRetry();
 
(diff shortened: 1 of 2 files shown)


jensotto:
I doubt that this will work as SharePoint will require the field to exist when the calculated field is created. This means that the PnP provisioning will validate and correctly transform the token into the correct display name even if the referenced field is not yet created, but SharePoint will throw an exception when it can't find the field in the site (or list) when the provisioning engine creates the field.
Please have a look at PR #817 for a possible implementation that will improve the extraction so that this scenario will be less likely to happen in the future. I have reordered the extracted fields so that the calculated fields are added to the provisioning template as the last site fields.

SchauDK:
@jensotto You're right. I completely messed up my test.
SharePoint will throw an exception it the referenced field doesn't exist.

PR #805, opened 2016-10-14
Fixed bug in ExportAllTerms where subterms were improperly formatted.…

casselc:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| New sample? | no |
| Related issues? | referenced in OfficeDev/PnP-PowerShell#476 |

For term sets containing terms with subterms, `ExportAllTerms` would incorrectly format the output as 
`Group|Group|TermSet|Group|TermSet|Term|SubTerm` instead of `Group|TermSet|Term|SubTerm`

This PR fixes that issue and changes the logic of `ExportAllTerms` to more closely match the logic and style of `ExportTermSet`. Also fixes a minor bug in `ParseSubTerms` where the method would recurse unnecessarily if the current subterm had no child terms.


jansenbe:
This PR can't be merged because the branch containing the fix in @casselc 's forked repository is not available. Can you please resubmit this PR again, but then leave the branch available until the PR is merged. 

Thanks.

Issue #496, opened 2016-03-26
[Provisioning Engine]: Allow Path in Files Src property.


MarkStokes-outlook:
This may not be a Provisioning Schema issue, but a Provisioning Engine issue, so I will duplicate this issue in both locations.

I would like to be able to add a local path to a file that needs to be provisioned by the Files section.
https://github.com/OfficeDev/PnP-Provisioning-Schema/blob/master/ProvisioningSchema-2015-12.md#file

Since the provisioning engine can "provision" Files into our site I would like to be able to structure my provisioning folder to replicate this structure.  For example, I would like to create a local folder called "_catalogs/masterpage" for my master pages and page layouts, then create local folders at "SiteAssets/Images", SiteAssets/css", "Pages", "Documents", "Pictures/Banners", etc.

This will help me to manage my deployable files, rather than a single flat file structure that will likely breakdown if I have files of the same filename that exist in different folders of my site.

At present, if I put a path in the Src property I get an error message that path names are not allowed, only filenames.

luismanez:
I've been using relative paths in the SRC attribute for a while:
```xml
        <pnp:File Src="..\Assets\Scripts\jquery.js" Folder="{site}/Style%20Library/js/" Overwrite="true" />
```

not sure about Full paths though, but works fine with relative paths. Just ensure you set the Connector of the Template, pointing to any folder, and then, the relative paths are relative to that folder.

jbranders:
@luismanez Can you give an example of your directory on the file system and what must i exactly do with the Connector? How can i let it point to any folder?

luismanez:
Something like this:

+pnp
++ProvisioningTemplate.xml

+src
++ Style Library
+++MyProject
++++Scripts
+++++Project.Core.js

And when you load the template and Apply it:

```c#
XMLFileSystemTemplateProvider provider = new XMLFileSystemTemplateProvider("Path_To_Pnp_folder", "");

var template = provider.GetTemplate("ProvisioningTemplate.xml");
template.Connector = new FileSystemConnector("Path_To_Pnp_folder", "");
```

and then, in the PnP XML file:

```xml
<pnp:File Src="..\Style Library\MyProject\Scripts\Project.Core.js" Folder="{site}/Style%20Library/myproject/scripts" Overwrite="true" />
```

If you are pointing your connector to the PnP folder, then the Src should be relative to that PnP folder.

Hope it helps.

bartdonninger:
Hello,

[...]

PR #807, opened 2016-10-17
Support both "\" and "/" as path separators in SharePointConnector

artokai:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | yes |
| New sample? | no |
| Related issues? | mentioned in #496 |

Added support for backslash (`\`) as a path separator to the PnP Provisioning Engine's `SharePointConnector`. 

Files which contain paths in their `Src`-attribute cannot currently be imported using the `SharePointConnector` because the Files-handler always adds a  backslash (`\`) in the path if the `Src`-attrbute contains folders. ([source](https://github.com/OfficeDev/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs#L440)). As the current `SharePointConnector` supports only forward slash (`/`) as the path separator, it causes an error.

Adding support for backslashes (`\`) in `SharePointConnector` fixes this problem and also makes it easier to use the same template both using the `SharePointConnector` and `FileSystemConnector` since both connectors now support the same separator character (backslash).

In the pull request I replace all backslashes to forwards slashes in all public methods, constructors and properties of the `SharePointConnector`. This ensures that the connector will work correctly even if the passed in container parameter is mixing both forward and backward slashes in the path. (This path with mixed separators results for example from the hardcoded backslash-separator in the `ObjectFiles`-handler)

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/Connectors/SharePointConnector.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/Connectors/SharePointConnector.cs
@@ -53,6 +53,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.Connectors
             {
                 throw new ArgumentException("container");
             }
+            container = container.Replace('\\', '/');
 
             this.AddParameter(CLIENTCONTEXT, clientContext);
             this.AddParameterAsString(CONNECTIONSTRING, connectionString);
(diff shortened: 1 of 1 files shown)


jansenbe:
Nice one  Arto :-) 

This indeed will increase usability, will merge. Thanks!

PR #783, opened 2016-10-05
 Implemented file level handling (draft/published) when provisioning …

jensotto:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| New sample? | no |
#### What's in this Pull Request?

 Implemented file level handling (draft/published) when provisioning files using provisioning templates

PR #815, opened 2016-10-21
File level implementation

teroarvola:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | -

Implemented setting file publishing level (Checkout/Draft/Published) in file provisioning.
 
Microsoft.SharePoint.Client.FileLevel string values map to OfficeDevPnP.Core.Framework.Provisioning.Model.FileLevel, is the PnP enum really needed? I think it could be replaced everywhere with Microsoft.SharePoint.Client.FileLevel, but didn't have time to do it now.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs
@@ -119,11 +119,8 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                             }
                         }
 
-                        if (checkedOut)
-                        {
-                            targetFile.CheckIn("", CheckinType.MajorCheckIn);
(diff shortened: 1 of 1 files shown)


jensotto:
Hi, this improvement has already been proposed in #783 

I guess the reasoning behind the two enums are to isolate changes. By having a PnP file level enum changes to Microsoft.SharePoint.Client.FileLevel will not affect the PnP schema.
This is something that you break when converting the two enums using Enum.Parse. Even though the chances are remote you open up for a runtime exception. This can happen if a new field level is introduced in PnP or it is changed in SharePoint. 

There is however one issue with the Checkout file level value. I'm not sure why it's allowed because it does not allow other users to check this in. This would make the user that provision the file the only one that could check in the file (unless I'm missing something). Your implementation would change the previous implementation to allow this scenario. That's why I have left the original logic inside the default case so that the file is always checked in.

teroarvola:
Good points, thanks for info.

PR #817, opened 2016-10-21
Fixed bugs when extracting and provisioning calculated fields.

jensotto:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| New sample? | no |
| Related issues? | fixes/improves #230, #537, #744, and #793, mentioned in #432 and attempted fixed in #466 |

Improved provisioning of calculated fields. One scenario is not fixed as tokens does not support this scenario:
CanProvisionCalculatedFieldLocallyInListInstance
This is when a calculated field references a field created directly in the list, and does not exist as a site field. This scenario is not working in the currently released version either. 

Test cases has been added to verify the functionality. These tests fails without this implementation:
CanUpdateCalculatedField
CanUpdateCalculatedFieldRefInListInstance
CanProvisionCalculatedFieldRefInListInstance

Changed site field extraction to move calculated fields to the bottom of the provisioning template site field section in order to ensure referenced fields are provisioned first. This does not take into account calculated fields that might reference other calculated fields.
This should be fixed, but I did not have time to look into that. The implementation should at least result in fewer issues than before.

The field extraction order also needs to be implemented for lists. But as long as it currently is not supported to use calculated fields with fields created in the list directly (not site fields) I have not looked into that.

The changes has been implemented based on findings by investigating implementation of SP2013 server side code in the methods Microsoft.SharePoint.SPList.ProvisionField (Microsoft.SharePoint.dll) and Microsoft.SharePoint.ApplicationPages.ManageFieldPage.UpdateFieldWithPropertiesFromXml (Microsoft.SharePoint.ApplicationPages.dll)

The second commit includes improved algorithm for transforming the formula during field extraction. This version eliminates the issue reported in #793.

Extended the GetFieldById and GetFieldByName methods to support returning the actual type and not force casting.

Regards
Jens Otto


jensotto:
The second commit includes improved algorithm for transforming the formula during field extraction. This version eliminates the issue reported in #793.

I do think that we also need to investigate how DefaultFormula and ValidationFormula will behave when extracting fields. I have not had time to investigate that now, but the changes to the calculated fields formula should at least now make extraction and provisioning of calculated fields more robust.

jansenbe:
Really nice work, like the fact that you also provided (updated) test cases :-) 

Will test and if ok merge this one. Thanks Jens!

PR #822, opened 2016-10-25
Bugfix : List instance datarows did not work if list instance's title…

SPParseError:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| New sample? | no |
| Related issues? |  |

There was a bug on datarows provisioning in the provisioning framework. Happening if list instance's title was localized (e.g. {resource:ListTitle}). The datarows provisining did get the list by title, using unparsed XML value. List {resource:ListTitle} was then not found and data rows could not be provisioned.


jansenbe:
Nice catch. Will test and merge. Thanks a lot!

PR #847, opened 2016-11-17
Added localization support for web part titles

wobba:
| Q               | A
| --------------- | ---
| New feature?    | yes

Added support to localize web part titles using resx files utilizing the approach mentioned by Alexey Sadomov at [http://sadomovalex.blogspot.no/2015/09/localize-web-part-titles-via-client.html](http://sadomovalex.blogspot.no/2015/09/localize-web-part-titles-via-client.html)

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/Functional/LocalizationTest.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/Functional/LocalizationTest.cs
@@ -1,16 +1,11 @@
 ﻿using Microsoft.SharePoint.Client;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using OfficeDevPnP.Core.Enums;
 using OfficeDevPnP.Core.Framework.Provisioning.Connectors;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
 using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers;
 using OfficeDevPnP.Core.Tests.Framework.Functional.Validators;
 using System;
-using System.Collections;
-using System.Collections.Generic;
 using System.Linq;
-using System.Xml.Linq;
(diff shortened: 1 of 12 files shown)


jansenbe:
Nice implementation Michael, including updated tests...me=happy :-)

We'll test and merge this one. Thanks!

jansenbe:
Hi Mikael,

Overnight I saw some of our test cases fail for a subset of test environments...further research showed that the used approach for adding multi-lingual labels only works when the user that's adding the translations has not explicitly set a language in their profile (like shown below):
image

Problem is that SharePoint will not allow to dynamically switch languages based on the header value and therefore the last added translation is the one that sticks for all languages.

I've adjusted the code to skip updating the web part titles when we're running under user context and the user has a preferred language configured.

wobba:
Hi Bert,
That makes total sense and an important detail indeed. Might be smart to add a warning as well if the user has a language set and are using wp tokens. I'll be using app tokens most likely, but my test user didn't have a default - so I didn't catch this.

Gotta love testing :)

wobba:
@jansenbe Out of curiosity can you explain the `if !SP2016` line in https://github.com/OfficeDev/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/Extensions/WebPartLocalizationExtensions.cs ? Don't you need to check the wp zone for all versions to make sure you fine the right part?

erwinvanhunen:
ZoneID is not supported on all versions unfortunately.

wobba:
So how do you position a wp in 2016? Wouldn't it be the same as in SPO, assuming the SP2016 flag is for 2016 onprem(?). Or is it the CSOM versions not supporting it (yet)?

erwinvanhunen:
It's the server side part of CSOM that is not supportjng it (yet).

PR #848, opened 2016-11-17
Set site collection locale to English in tests

wobba:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

If the default site collection uses tenant language, for example Norwegian, then localization tests will fail as they assume 1033/English to be the default.

I also set the the initial quota for storage to 100mb and sandbox resources to 0, to play nice.


jansenbe:
Thank, we'll process this one together with your other PR.

PR #859, opened 2016-11-26
Fix in list instance provisioning : list title and list url were not …

SPParseError:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | no

Fixed error when provisioning list instances with localized titles and urls : those properties were not parsed when searching for an existing list, making code fail after when trying to create it (a list already exists).


jansenbe:
Thx Stéphane, this change makes sense and we'll merge it.

PR #867, opened 2016-11-30
Token is not replaced for a SecurableObjectExtensions.RoleDefinition in SetSecurity

bartdonninger:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | none

A parameter/token cannot be used in the RoleDefinition of a RoleAssignment. The parse.ParseString method wasn't used when trying to retrieve the RoleDefinition on the Web. It will not retrieve a RoleDefinition and will produce an error when the RoleDefinitionCollection is empty when adding it to the RoleAssignment.


jansenbe:
Thanks for this change Bart, looks good and we'll process it

PR #884, opened 2016-12-07
Fix: WelcomePage can now be overwritten when tokens are used to build urls

sebastianmattar:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no 
| Related issues?  | -

The provisioning framework failed when a Page that was set as a WelcomePage was to be overwritten and tokens were used to build the Page URL. 
I updated the comparison mechanism to to detect if the current page is the WelcomePage, removed some deprecated references and changed the code that updates the WelcomePage property after the Page was provisioned.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectPages.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectPages.cs
@@ -26,6 +26,14 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 
                 web.EnsureProperties(w => w.ServerRelativeUrl, w => w.RootFolder.WelcomePage);
 
+                string welcomePageUrl = web.RootFolder.WelcomePage;
+
+                string welcomePageServerRelativeUrl = null;
+                if (!string.IsNullOrEmpty(welcomePageUrl))
+                {
+                    welcomePageServerRelativeUrl = UrlUtility.Combine(web.ServerRelativeUrl, welcomePageUrl);
+                }
+
                 // Check if this is not a noscript site as we're not allowed to update some properties
                 bool isNoScriptSite = web.IsNoScriptSite();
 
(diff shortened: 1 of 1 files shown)


jansenbe:
Thx Sebestian, we'll be looking into this change.

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectPages.cs
```diff
@@ -61,8 +61,13 @@ public override TokenParser ProvisionObjects(Web web, ProvisioningTemplate templ
                             {
                                 scope.LogDebug(CoreResources.Provisioning_ObjectHandlers_Pages_Overwriting_existing_page__0_, url);
 
-                                if (page.WelcomePage && url.Contains(web.RootFolder.WelcomePage))
+                                string welcomePageUrl = UrlUtility.Combine(web.ServerRelativeUrl, page.ParentTemplate.WebSettings.WelcomePage);
```
sebastianmattar: Hi Bert,

thank you for the feedback - I still have to get a deeper understanding on how everything works so please bear with me ;)

You are correct, I should add a check if the welcome page is defined. 

As I want to consider update scenarios (where the template does not necessarily define WelcomePage although the web has one set) I don't want to rely on the web settings handler to set the WelcomePage back to the correct value.

jansenbe:
Hi Sebastian,

I've added some inline comments to your changes. This PR in current format will break many customers cause Pages and WebSettings handlers are not depending on each other and we want to keep this as well. Does the proposed approach in the code comments not work for you?

sebastianmattar:
Hi Bert,
I can not see your inline comments. Nevertheless I removed the dependency on the WebSettings object. Could you have a look?

jansenbe:
Hi Sebastian,

The welcome page handling in the page objecthandler has been deprecated a while ago. It's advised you use the websettings object handler to set your welcome page. As such I'm not merging this PR, let me know if you can't realize your scenario using the welcome page handler.

PR #886, opened 2016-12-07
Updated the GetPagesLibrary extensions method.

bartdonninger:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no 
| Related issues?  | -

Updated the GetPagesLibrary to get the PublishingPage library by using the PagesListId from the web propertybag entries.

In the previous code the cmscore resources file was used to retrieve PagesListName. I am using a Dutch publishing site and the translation value in the cmscore is wrong. So the PagesList could not be retrieved. 

I went and had a look in the Microsoft.Sharepoint.Publishing.PublishingWeb DLL to see how it retrieved the pages library in there. It also uses the __PagesListId web propertybag entry. 

So i think it is better to also rely on that property instead of reading it from a resources file.

### DIFF
--- a/Core/OfficeDevPnP.Core/AppModelExtensions/ListExtensions.cs
+++ b/Core/OfficeDevPnP.Core/AppModelExtensions/ListExtensions.cs
@@ -764,19 +764,31 @@ namespace Microsoft.SharePoint.Client
         {
             if (web == null) throw new ArgumentNullException(nameof(web));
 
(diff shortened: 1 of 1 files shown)


jansenbe:
Thanks for this PR Bart. I've looked at the issue and have done a slightly different implementation which introduces no potential change in behavior. We'll get you the needed credits for the idea, so thanks for bringing this to our attention.

PR #888, opened 2016-12-07
Added localization support to list views

wobba:
| Q               | A
| --------------- | ---
| New feature?    | yes

Added support to localize list view titles, with same approach as used for web part titles.

Renamed WebPartLocalizationExtensions to LocalizationExtensions.


jansenbe:
Thanks again Mikael. Good addition to the work already done in this area!

PR #895, opened 2016-12-10
Re index group point pub

wobba:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

Some sites in SPO no longer allow access to Web property bag, so for re-indexing we need to trigger individual lists.

Added check on NoCrawl flag for web/list as there is no need to try to trigger these.


jansenbe:
Thanks Mikael, again a good update!

PR #909, opened 2016-12-24
Template creation: Add some automatic fields that don't need to be referenced

sebastianmattar:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no 
| New sample?      | no 
| Related issues?  | -

I created a template from a migrated SharePoint 2010 site collection. I found that the template contained some field references to automatic fields that caused errors during provisioning.

I seems like old site collections can use different names for some out of the box fields. 
As a quick fix I added the names that were in my site collection to the code. In the long run we should use the Field.ID property to identify those fields by their Guid.


jansenbe:
Thx Sebastian, this is a good addition which I'll be merging

PR #940, opened 2017-01-05
Added an optional function in certain tenant methods to allow for cancellation

erwinvanhunen:


erwinvanhunen:
Very true. It's a minor change, I'm coding it now.

erwinvanhunen:
See PR #941

artokai:
This is better, but it does not work... 

You return true if WaitForIsComplete is cancelled. And the CreateSiteCollection function seems to think that "true" means that the site collection has been succesfully created. 

I think it would make more sense if the WaitForIsComplete would return false, when the wait is cancelled. This would make it harder to make these kinds of mistakes.

erwinvanhunen:
We're getting there :-) PR #944 . Indeed, it made more sense to have WaitForIsComplete to return a false if it was cancelled.

artokai:
Almost there, but not yet =)

The last thing to do fix to do is to make the `WaitForIsComplete` to return `false` when it has been timed out. Now it returns`true`for all of it code paths.

https://github.com/erwinvanhunen/PnP-Sites-Core/blob/950a336838e990be3fd8e3a1f7958ebe0da52445/Core/OfficeDevPnP.Core/AppModelExtensions/TenantExtensions.cs#L698

artokai:
Submitted a small pull request #956 to return `false` from `WaitForIsComplete` when it times out.

erwinvanhunen:
Merged. We had to create a .1 release as we -just- released the Core build out to Nuget. But it's in there now. Thanks!

Issue #936, opened 2017-01-05
Error running multiple Jobs from the same windows service or console app


emilverwoerd:
**Category**
[x ] Bug
[ ] Enhancement

 **Environment**
[ x] Office 365 / SharePoint Online
[ x] SharePoint 2016
[ x] SharePoint 2013

**Expected or Desired Behavior**
Allows other jobs to continue when multiple jobs are running in the same console application or windows service.

**Observed Behavior**
On my development machine I have a windows service that runs multiple timerjobs in the same thread at once. However when a job finishes the Timerjob signals all others that it is done and so the console or windows service will discard the process.

The fix is replacing the following code in the TimerJob.cs class
   // Threading
        private static int numberOfThreadsNotYetCompleted;
        private static ManualResetEvent doneEvent;

by
     // Threading
        private int numberOfThreadsNotYetCompleted;
        private ManualResetEvent doneEvent;

jansenbe:
Hi @emilverwoerd ,

This change has been implemented. Closing this issue now.

PR #942, opened 2017-01-05
Replaces static for normal variables for Timerjob #936

emilverwoerd:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | Issue 936

The timerjob class is fixed to not use static variables which gives problems when using multiple timerjobs in a single windows service or console application


jansenbe:
Hey Emil,

Thanks for this change. Looks good and we'll merge it. Thx;

emilverwoerd:
Thx when the February release is there we can use our new Windows Service for all the custom jobs we built :-)

PR #946, opened 2017-01-06
Added missing checks for null value when creating a term.

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

When adding terms that are reused but not the source a null reference exception is thrown because the CreateTerm method will return null in this scenario. there are 5 references to that method and only three has a null check implemented for the return value. I have added this check to the remaining 2 references.


PaoloPia:
Hi Jens,
Good catch! Thanks for your contribution.

Issue #930, opened 2017-01-04
Enhancement request: Support cancelling / timing out long running tasks


artokai:
#### Category
[ ] Bug
[X] Enhancement

#### Environment
[X] Office 365 / SharePoint Online
[X] SharePoint 2016
[X] SharePoint 2013

#### Description

The TenantExtensions allows us to easily create new site collections and wait for the site collection job to finish. This is great, but sometimes there is something weird going on at the tenant and the site collection creation can take multiple hours to finish. If for example an Azure WebJob is waiting for the site collection creation job to finish, then the webjob will get "stuck".

It would be great if the PnP library would allow us to set a timeout value for these long operations. If a site collection would not get created in a timely fashion then PnP would exit the waiting loop and the caller could report a possible error to the end users.

One way to implement this would be to allow the caller to pass a [CancellationToken](https://msdn.microsoft.com/en-us/library/dd997364(v=vs.110).aspx) to the long running operations in PnP.  If the caller would then cancel the token, then PnP could notice this and exit the waiting loop. This approach would work well with Azure WebJobs since they support [`TimeOut`-attribute](https://github.com/Azure/azure-webjobs-sdk-samples/blob/master/BasicSamples/MiscOperations/Functions.cs#L125) which will automatically cancel any long running jobs when the timeout-value is exceeded.

What do you think about supporting timeouts / cancelling? If you think that this kind of functionality can be added to the long running operations in the TenantExtensions, then I'm happy to create a pull request for it.

erwinvanhunen:
Hi!

I just merged an update.

The Tenant methods (long running ones) now allow for an optional Func<TenantOperationMessage, bool> parameter. This function will be called during polling for the status of the long running operation. If you return true from this message, it will cancel the wait loop.

TenantOperationMessage is an enum which can help you to figure out what's going on when the function is being called.

An example

```csharp
Tenant.CreateSiteCollection(mySiteentity, timeoutFunction: (operationmessage) =>
            {
                if (operationmessage == TenantOperationMessage.CreatingSiteCollection)
                {
                    // write out some progress
                }
                return false; // set to true to cancel the wait loop
            });
```

PR #956, opened 2017-01-09
Bugfix: WaitForIsComplete should return false when timed out

artokai:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | #930

Fix for a bug in the new TimeOut-feature for long running operations in TenantExtensions. The WaitForIsComplete-function should return false, when timed out.

Discussion about this feature and bug can be found here:
https://github.com/SharePoint/PnP-Sites-Core/pull/940#issuecomment-270871238


erwinvanhunen:
Good catch!

PR #958, opened 2017-01-09
Replace Microsoft.IdentityModel.Extensions with SharePointPnP.IdentityModel.Extensions 

SharePointRadi:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | not really
| New sample?      | no
| Related issues?  | Replaced Microsoft.IdentityModel.Extensions with SharePointPnP.IdentityModel.Extensions

We don't want to depend on Microsoft.IdentityModel.Extensions, so we created our own SharePointPnP.IdentityModel.Extensions assembly. This commit removes the reference in Core and adds a reference to our Release build of SharePointPnP.IdentityModel.Extensions

Note: I've made sure I don't update the target Framework version to 4.5.1, we should do one thing at a time. Here I'm keeping it for v4.5


jansenbe:
Thx Radi. Will have a look and merge. We're still in time for this nights test runs

PR #977, opened 2017-01-16
Check for empty and white space and not just null

wobba:
If not, code breaks in the WriteMessage parser

| Q               | A
| --------------- | ---
| Bug fix?        | yes

Bug fix for formatting the refactored Message line in ObjectField


erwinvanhunen:
good catch!

PR #986, opened 2017-01-20
Implement *smart* WillProvision for ObjectLookupField

dbeaulieu:
| Q               | A
| --------------- | ---
| Bug fix?        | Yes
| New feature?    | No
| New sample?      | No
| Related issues?  | N/A

The WillProvision method for ObjectLookupField was always returning true.  This would in turn always call ProvisionObjects and this would always execute queries against SharePoint before determining if the template had elements applicable to provisioning Lookup Fields.  This change fixes the performance *bug*.
1. Changes WillProvision to only return true if the template has lookup fields in SiteFields or in Lists.
2. Creates a helper property in Field to avoid multiple parsing of SchemaXml (used by WillProvision and ProvisionObjects).  This property returns the parsed XElement from the SchemaXml string and will be re-parsed whenever the underlying SchemaXml changes.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/Model/Field.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/Model/Field.cs
@@ -24,7 +24,11 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.Model
         public string SchemaXml
         {
             get { return this._schemaXml; }
-            set { this._schemaXml = value; }
+            set
+            {
+                this._xelement = null;
+                this._schemaXml = value;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

dbeaulieu:
@jansenbe I don't see the changes on SharePoint:dev, i'm wondering if the PR I submitted for #1004 interfered with this one.

jansenbe:
Interesting...somehow git decided to mark this PR as merged because I merged #1004...seems like a bug to me. Anyhow, we discussed this PR today and our feedback was to not merge this one since the performance impact is very small.

dbeaulieu:
Ok.  The use case I was attempting to optimize is deploying the same template to many sites, so every bit would help, although I have not benchmarked.

PR #1071, opened 2017-02-23
Handle EnsureUser exceptions

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no?
| New sample?      | no?
| Related issues?  | NA

When exporting site security the template cannot be applied to a new site collection if users are disabed in Azure AD.

Eg. a user within a site group is disabled because he left the company. When applying the template the site groups are created and the users are added to the group. However when the framework is trying to add the disabled user, EnsureUser will throw an exception and stop further progress.

This change logs a Warning instead and continues the progress.

EnsureUser will also fail for External Users if they don't already exist on the site.
A check has been added skipping external user if doesn't exist in SiteUsers.


jansenbe:
Hi Michael, definitely a good change which I'll process.

PR #1083, opened 2017-02-27
Re-order fields when content type is updated

SuleymanA:
| Q               | A
| --------------- | ---
| Bug fix?    |  yes
| New feature?    | no
| New sample?      | no 
| Related issues?  | not found

Updated the content type object handler to reorder the content type field after adding new fields to the content type. Re-order was called too early so new fields added to the content type always ended up at the bottom. Reduced to one re-order call.


jansenbe:
Good catch @SuleymanA , will merge this one.

PR #1086, opened 2017-02-28
Ensure sealed content type is not updated

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

This PR fixes an issue where Sealed content types are updated.
This change only allows sealed content types to updated if they're changed not to be sealed.


jansenbe:
Makes sense to not fail on this one, thanks for the fix

PR #1097, opened 2017-03-06
Only add labels to a term for available termstore languages

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Only add labels to a term, if the label language is an available termstore language.


jansenbe:
Again a good change, thanks for that!

PR #1101, opened 2017-03-07
Optimize busy wait and better error handling

wobba:
| Q               | A
| --------------- | ---
| Improvement?        | yes

* Changed Thread.Sleep to Task.Await to avoid busy sleep
* Added try/catch blocks to get Graph errors out in the error stream since we are doing async calls


jansenbe:
Thanks Mikael. Will have a look at this one.

PR #1109, opened 2017-03-09
Added try/catch with Logging on all calls

wobba:
| Q               | A
| --------------- | ---
| Improvement?        | yes

Added try/catch on all public functions to log service errors. Very useful when using PnP posh, as you can more clearly see what the actual error is.

`C:\Users\miksv> New-PnPUnifiedGroup -DisplayName "Foo" -Description "Foo" -MailNickname footest -Owners miksvenson@tenant.onmicrosoft.com

powershell.exe Error: 0 : 2017-03-09 12.55.40.7791      [OfficeDevPnP.Core]     [0]     [Error] Graph call returned the following error: One or more properties contains invalid values..   0ms`


jansenbe:
Good one, just in time for our master merge of tomorrow :-)

PR #1129, opened 2017-03-16
default conversion to string

ivanvagunin:
Convert source value to string if destination property is of string type


PaoloPia:
Merged! Thank you!

ivanvagunin:
Thanks! Will continue with serializers then

PR #1143, opened 2017-03-27
Support switching between Article and Home layout on a modern page

wobba:
| Q               | A
| --------------- | ---
| New feature?    | yes

Support using both the modern page home page layout as well as the article layout. The home page layout has no header, and is very useful.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/PageExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/PageExtensions.cs
@@ -12,11 +12,6 @@ using OfficeDevPnP.Core.Utilities;
 using Microsoft.SharePoint.Client.Publishing.Navigation;
 using OfficeDevPnP.Core.Utilities.WebParts;
 using PersonalizationScope = Microsoft.SharePoint.Client.WebParts.PersonalizationScope;
-using System.Net;
-using System.IO;
-using System.Text;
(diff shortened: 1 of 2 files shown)


wobba:
Hold on on this. Need to add support so a Home page is created without ooh parts.

jansenbe:
Good stuff Mikael :-)

wobba:
@jansenbe Ok.. added support to keep oob wp's if the layout has any (like the home layout). Option to promote as news as well as set the Title field.

Let me know what you think. The code will allow the change of layout and promoted state - which should be ok I think.

wobba:
@jansenbe Did you get a chance to look at this?

jansenbe:
Hi @wobba,

Yes and no...I did look but did not take action cause I'll process this together with upcoming changes that I'll need to do as we're "improving" the modern pages. So stay tuned :-)

wobba:
Got it.. I'll have to wait before starting powershell dev then to see what it will look like :)

jansenbe:
Hi Mikael, as discussed I've not directly merged your PR but I've taken some of your code and added that as part of bigger refactoring that prepares the PnP engine to support the recently announced modern page capabilities like flexible zones and commenting. Closing this PR now, but we'll credit you  for the input on this one.

PR #1167, opened 2017-05-02
small fix of  NullReferenceException in CredentiaManager.GetSharePointOnlineCredential

rychlym:
here is public static method GetSharePointOnlineCredential  in the CredentiaManager (the first method from the top of the class ). It seems to be  not used inside of Library, but I allowed it to use. In case there is problem with credentials it crashes with NullReferenceException) 
Seeing   GetSharePointOnlineCredential  using  GetCredential method, which returns null  in case of no success, I allowed to ask for this simple pull request , where the GetSharePointOnlineCredential  similarly returns null in case of no success.


jansenbe:
Hi @rychlym, this is a good change and we'll get this one merged. Thanks for your contribution!

PR #1204, opened 2017-06-05
Null exception in TokenParser class when having empty parameter in provisioning template

afsandeberg:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

When having empty parameter values in a provisioning template I get an exception from the TokenParser class as the empty value get translated to a null value and gives an error in the `regex.Replace` call. Converting null value to empty string at creation solves the issue.


jansenbe:
Thanks @afsandeberg, good catch!

PR #1205, opened 2017-06-06
Issue with SiteCollection TermSet and Terms

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | NA

Consider this scenario.
1. Export a SiteCollection Term and TermSet and assign the exported template to a variable
2. Apply the template to a Site Collection
3. Apply the same template to another Site Collection
You now get the error "Failed to read from or write to database".

The issue is that Term and TermSet is assigned a Id. When the TermSet and Term are created the second site collection a TermSet and Term ids already exist in the database. 

This fix assigns a new id to the TermSet if the TermGroup is SiteCollection.
Also if a Term isn't a reused Term a new Id is assigned.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTermGroups.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTermGroups.cs
@@ -148,10 +148,16 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                         }
                         if (set == null)
                         {
+                            if (modelTermGroup.SiteCollectionTermGroup)
+                            {
+                                modelTermSet.Id = Guid.NewGuid();
+                            }
+
                             if (modelTermSet.Id == Guid.Empty)
                             {
                                 modelTermSet.Id = Guid.NewGuid();
                             }
+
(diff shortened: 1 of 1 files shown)


jansenbe:
Thx for this change @SchauDK. We'll be discussing this one first as you change is removing the ability to provision termsets and terms with a given ID across environments.

jansenbe:
Hi Michael, can you improve this PR so that the default behavior stays the same and as such the engine still can provision termsets and terms with "known" id's. Possible implementation (for site collection scoped term(set) would be to:
-define a list of used term(set) id's
-check for existence of term(set) with the given id by checking the list first, if not there check the termstore
-if yes, store that id in a list for future retrieval and generate a new id 
-if no, use the id that was provided in the template

Other ideas are welcome as well.

Alternative solution for you can be to remove the id's from the generated template.

SchauDK:
Sure, I can improve this PR.
I'm actually doing the alternative solution you're suggesting, but I think it would be nicer if the framework is handling it.

jansenbe:
Hi Michael,

I'm closing this PR as there's no blocking issue + there's an alternative solution. Feel free to create a new PR with the described alternative approach if this is a key issue for you.

PR #1211, opened 2017-06-09
Update {listviewid:...} token

wazawski:
According to source code, the token {listviewid:...} should actually be {viewid:...}. I just spend more than an hour to figure out why my webpart couldn't be provisioned.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| New sample?      | no

The file ProvisioningEngineTokens.md was updated by the correct token to get a view id : {viewid:...} instead of {listviewid:...}


jansenbe:
Hi @wazawski, thanks for this change. I've manually merged this to our dev branch since you targeted our master branch. We normally only accept changes on our dev branch, but since this was a small change we processed it anyhow.

PR #1214, opened 2017-06-12
Support modifying Forms pages by null-checking ListItemAllfields

okms:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | none

When treating files its a bit dangerous to assume all files live in lists. One example of this is when modifying existing forms pages, e.g. adding webparts to these. For us this is a pretty common scenario, we often add a content editor webpart or script editor webpart to forms pages so we can modify OOB functionaltity without creating custom Forms pages.

In order to support adding webparts to pages which are NOT normal files, we have to do a null check on listitemallfields whenever we use that.


jansenbe:
Thx Ole, good change which we'll merge in.

PR #1217, opened 2017-06-14
Handling duplicated field reference causing formula syntax error for …

litex1982:
…calucated field

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

When a calucated field contains formula referring to one field for mutliple times, formula =IF(NOT(ISBLANK(Region)),Title&" - "&[Project No.]&" ("&Region&")","") which is referring to Region twice and thus region field will be listed twice in the FieldRef and in thise case the current implementation will try to replace Region twice and leads to incorrect syntax in formula


jansenbe:
Hi @litex1982, thanks for you contribution. Before further processing this I want to understand why you are using this deprecated method versus the one we are using internally (ObjectField.TokenizeFieldFormula). Inside the PnP engine we are not using the method you've fixed, it's only there for backwards compatibility.

PR #1222, opened 2017-06-19
#1208: filter content types that come with content type syndication

### LINKED ISSUE #1208: GetProvisioningTemplate: add ability to exclude fields and content types that came from content type hub
#### Category
* Enhancement

#### Environment
[x] SharePoint 2016, May 2017 CU
[ ] SharePoint 2013

_Not environment-dependent_

#### Observed Behavior
Templates exported with ([GetProvisioningTemplate](https://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Extensions/WebExtensions.cs)) contain fields and content types that came from content type hub, and there is no way to exclude them with [ProvisioningTemplateCreationInformation](https://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ProvisioningTemplateCreationInformation.cs) properties or [Handlers](https://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Provisioning/Model/Handlers.cs).

Global content types and fields misrepresent actual site configuration and can create conflicts during site configuration migration between site collections , web applications and farms (e.g. unable to override read only content types).

Add a way to include/exclude fields and content types that came from Content Type Hub. For example
 * Do not export fields and content types that came from content type hub by default
 * Add Boolean property `IncludeAllFields`
 * Add Boolean property `IncludeAllContentTypes`

Global fields and content types would have  property `SourceId` = `ContentTypeHub.WebId` 

#### Steps to Reproduce
* Create a few fields on the Content type Hub site
* Use these fields in a few new content types
* Publish the content types
* Wait until the content types appear in the subscriber sites (or, run Publish and Subscribe timer jobs)
* Add a few fields and content types to the subscriber site.
* Export subscriber site template, e.g.
```powershell
Get-PnPProvisioningTemplate -Out "subscriber-site.export.xml"
```
* Notice that the fields and content types created in the content type hub appear in the template along with the subscriber site only fields and content types.
* Make a few changes in the Content Type Hub site fields and content types (e.g. change descriptions) and re-publish the content types
* Create another subscriber site and provision the exported template to the new site
  * Most likely, you are going to experience errors during import when the engine is going to try updating read only fields and content types that came from the content type hub site.

stevebeauge:
| Q               | A
| --------------- | ---
| Bug fix?        | no 
| New feature?    |  yes
| New sample?      | no 
| Related issues?  | fixes #1208

When a content type is provisioned through content type syndication, it's flagged as *read-only* in the site collection.

This leads to provisioning error because the PNP provisioning engine try to create the content type, which is already here because of the ct syndication.

My solution was simply to check if the content type is syndicated or not, to avoid exporting the content type.

A flag has also been added to the `ProvisioningTemplateCreationInformation` to reenable exporting this kind of content types.

Hope that helps

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectContentType.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectContentType.cs
@@ -10,6 +10,10 @@ using OfficeDevPnP.Core.Framework.Provisioning.Connectors;
 using System.IO;
 using System.Text.RegularExpressions;
 using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.TokenDefinitions;
+using System.Xml.Linq;
+using System.Xml.XPath;
+using System.Xml;
+using System.Text;
 
 namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 {
@@ -476,7 +480,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
         private IEnumerable<ContentType> GetEntities(Web web, PnPMonitoredScope scope, ProvisioningTemplateCreationInformation creationInfo, ProvisioningTemplate template)
         {
             var cts = web.ContentTypes;
-            web.Context.Load(cts, ctCollection => ctCollection.IncludeWithDefaultProperties(ct => ct.FieldLinks));
+            web.Context.Load(cts, ctCollection => ctCollection.IncludeWithDefaultProperties(ct => ct.FieldLinks, ct=>ct.SchemaXmlWithResourceTokens));
             web.Context.ExecuteQueryRetry();
 
             if (cts.Count > 0 && web.IsSubSite())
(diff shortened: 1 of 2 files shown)


stevebeauge:
I've signed the cla... not sure if it's supposed to auto update the labels (if it is, there may be a problem), or if it's a manual process.

jansenbe:
Hi @stevebeauge, thanks for this contribution. The changes you've done makes a lot of sense and I've merged it but did a small, but important, change: in order to stay backwards compatible I've changed your default from excluding CTHub CT's to include them. You can obviously control this via the creation params.

stevebeauge:
@wy62, I never had time to implement it, but as far as I remember fields' sourceid attribute is set to the web ID.
Probably fields with sourceid neither equal to web's id nor any of web 's ancestor (up to site coll root) nor the const http://.... are from the CT hub.
It's only a quick guess, but it should work.

PR #1231, opened 2017-06-21
Fixed replacement of ListId with token to properly escape list title …

TheJeffer:
…so that it doesn't break when title contains an ampersand.

| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    |no
| New sample?     |no
| Related issues?  |none

Currently, if you try to use Get-PnPProvisioningTemplate on a site containing list view web parts on the home page, it will break if the list title contains an ampersand. This is because there are several places in the code where a find/replace is done directly on the xml to tokenize the list ID, but the list title value used in the token is not properly xml escaped.


jansenbe:
Hi Jeff,

Good catch and thanks for your contribution to pnp. Can you sign the CLA so that we can process this PR.

TheJeffer:
Absolutely. Just waiting on approval from my boss to get that signed.

TheJeffer:
CLA has been signed. Thanks!

PR #1240, opened 2017-06-28
Fixed to properly detokenize xml strings containing xml escaped token…

TheJeffer:
Fixed to properly detokenize xml strings containing xml escaped tokens.

Updated TokenParser.ParseString to be simpler, more performant, and safer (now prevents infinite loops in nested parameters).

Added simple unit test for testing nested parameters.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | https://github.com/SharePoint/PnP-Sites-Core/pull/1231

The TokenParser.ParseString method did not properly handle XML strings containing XML escaped tokens. The token regexes would not match a token that had been XML escaped. I've added a new method that converts the XML string into an XmlDocument, then searches through attributes and element text values, replacing tokens as necessary, then returns back the full XML string.

I've also Updated TokenParser.ParseString to be simpler, more performant, and safer (now prevents infinite loops in nested parameters). Previously, one could cause an infinite loop by using nested params where param A contains a token for param B which contains a token for param C which contains a token for param A. I also added a short-circuiting condition to skip searching a string for tokens if the string doesn't contain a { or ~ character so that we're not needlessly running hundreds of regex searches against a string that doesn't contain any tokens. I don't see a hard requirement for tokens to begin with one of those two characters, but it seems to be the convention for all the tokens currently constructed in this project. If that's not the case, feel free to remove that condition from the initial if statement in the ParseString method.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs
@@ -355,69 +355,57 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
         /// <returns>Returns parsed string</returns>
         public string ParseString(string input, params string[] tokensToSkip)
         {
-            var origInput = input;
-            if (!string.IsNullOrEmpty(input))
-            {
-                foreach (var token in _tokens)
-                {
-                    if (tokensToSkip != null)
-                    {
-                        var filteredTokens = token.GetTokens().Except(tokensToSkip, StringComparer.InvariantCultureIgnoreCase);
-                        if (filteredTokens.Any())
-                        {
-                            foreach (var filteredToken in filteredTokens)
-                            {
-                                var regex = token.GetRegexForToken(filteredToken);
-                                if (regex.IsMatch(input))
-                                {
-                                    input = regex.Replace(input, token.GetReplaceValue());
-                                }
-                            }
-                        }
-                    }
-                    else
-                    {
-                        var matchingTokens = token.GetRegex().Where(regex => regex.IsMatch(input));
-                        foreach (var regex in matchingTokens)
-                        {
-                            input = regex.Replace(input, token.GetReplaceValue());
-                        }
-                    }
-                }
-            }
+            if (string.IsNullOrEmpty(input) || input.IndexOfAny(new []{'{','~'}) == -1) return input;
 
-            while (origInput != input)
+            var tokensToSkipList = tokensToSkip?.ToList() ?? new List<string>();
+
+            foreach (var token in _tokens)
             {
-                foreach (var token in _tokens)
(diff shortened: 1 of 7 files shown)


Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs
```diff
@@ -355,69 +355,57 @@ public IEnumerable<string> GetLeftOverTokens(string input)
...
+                foreach (var node in nodes.OfType<System.Xml.XmlElement>())
+                {
+                    if (!string.IsNullOrEmpty(node.InnerText))
                     {
-                        foreach (var regex in token.GetRegex().Where(regex => regex.IsMatch(input)))
-                        {
-                            origInput = input;
-                            input = regex.Replace(input, token.GetReplaceValue());
-                        }
+                        node.InnerText = ParseString(node.InnerText, tokensToSkip);
```
SchauDK: node.InnerText = HttpUtility.HtmlEncode(ParseString(node.InnerText, tokensToSkip));

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs
```diff
@@ -355,69 +355,57 @@ public IEnumerable<string> GetLeftOverTokens(string input)
...
-                            foreach (var filteredToken in filteredTokens)
-                            {
-                                var regex = token.GetRegexForToken(filteredToken);
-                                if (regex.IsMatch(input))
-                                {
-                                    input = regex.Replace(input, token.GetReplaceValue());
-                                }
-                            }
-                        }
+                        attribute.Value = ParseString(attribute.Value, tokensToSkip);
```
SchauDK: attribute.Value = HttpUtility.HtmlEncode(ParseString(attribute.Value, tokensToSkip));

TheJeffer:
That shouldn't be necessary, since we're using an XmlDocument. The OuterXml property will already give back properly encoded xml. I put together a quick example to show that the values are encoded properly... https://dotnetfiddle.net/1id4dy.

SchauDK:
You're so right. No worries then.
Thanks for the explaination.

NicolajHedeager:
It looks like this change breaks parsing of termset ids for site collection termsets.

You can reproduce this error by adding a site collection term called 'My template set' and the try to parse the string: {termsetid:{sitecollectiontermgroupname}:My template set}

In the previous version of the ParseString method, this would return the id of the termset 'My template set', but the implementation of the method will only parse the {sitecollectiontermgroupname} token and not the termsetid token. This is because the 'while (origInput != input)' loop has been removed.

TheJeffer:
You're absolutely right, @NicolajHedeager. Missed that use case. Fixed the change and added a test for it.

jansenbe:
Hi @TheJeffer, a great change which will be merged. Thanks for your contribution and apologies for the delay in merging.

PR #1241, opened 2017-06-29
New Feature: Provision WikiPages into SubFolders of a Library

PascalStrassl:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | none

If you provide a path to a WikiPage, that contains a folder in the library, the Provisioning will fail.
e.g: /sitepages/SubFolderA/somePage.aspx

This change will ensure that the target folder (SubFolderA) is created, before the Wiki Page is uploaded.


jansenbe:
Thanks Pascal, good change. I've added this one as a test case for the pages object handler functional test so we keep track of future stability of this change.

PR #1324, opened 2017-09-06
Fix webpart importation

### LINKED ISSUE #869: Adding .DWP as part of page when provisioning template
Please refer to this Issue: https://github.com/SharePoint/PnP-Provisioning-Schema/issues/143

stevebeauge:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no 
| Related issues?  | fixes #869

`LimitedWebPartManager.AddWebPart` method from the CSOM library requires to provide WebPart definition wrapped within a `<webParts>` node.

The custom serialization reads the `pnp:webPart/pnp:Contents/webParts` **InnerXml** value, which is actually the `<webPart>` node.

This leads to the namespace error mentioned in the issue.

Changing the serializer code to use **OuterXml** resolves the issue.

This also allow the usage of DWP webparts.

Hope that helped


PaoloPia:
Thanks Steve,
That's a very good one! I'll merge it.

spcfran:
Nice catch @stevebeauge, been banging my head all day against this!

In case someone needs a quick workaround to make this work until the next release (as it is my case), just add an extra dummy xml tag under `pnp:Contents`:

image

stevebeauge:
The workaround can probably be automated when using the PowerShell  `Apply-PnpProvisoningTemplate` command.

I usually load the template using `Load-PnPTemplate`, tweak the template using a script and apply the template using `InputInstance` parameter instead of the `Path` parameter.

The script could be able to check the Contents property of each webPart to add the missing node.

PR #1333, opened 2017-09-12
If page does not have overwrite and exist, then continue in the loop

wobba:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

Fix is in line 64. If not you get null ref exception in line 74.

Also applied the Boy Scout Rule - made code cleaner with small refactorings and removing unused code.


jansenbe:
Good catch Mikael, will process this one

PR #1371, opened 2017-09-28
Refresh Token bug fix in WebAPI Helper

ypcode:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

Fixes the condition to refresh the access token in the WebAPI Helper
The Token was refreshed only after expiration+ one hour


ypcode:
Hello @jansenbe,
In some docs, I read the ACS access token was initial valid for 12 hours but this could vary.
In Azure AD tokens, I read it is valid for only one hour. I was wondering if it has become the same for ACS tokens, because that would mean this condition I fixed here is no longer relevant.
Any thoughts about it ?

jansenbe:
Think token lifetime (is) will be configurable, so this still makes sense.

PR #1374, opened 2017-09-29
Adds extension to register WebAPI in ASP.NET MVC solutions instead of using Page object

ypcode:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes

Basically copies the exsting RegisterWebAPIService() method but allows to use it with a HttpContextBase instance instead of a Page object not available in ASP.NET MVC (with Razor views) project


jansenbe:
Good addition

PR #1379, opened 2017-10-02
Allow site columns to be provisioned to subwebs when ProvisionContentTypesToSubWebs flag is set

TheJeffer:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

If the ProvisionContentTypesToSubWebs flag is set, the engine should allow provisioning of site columns to the subweb as well. Alternatively, if this needs to be more explicitly enabled, a new ProvisionSiteColumnsToSubWebs flag could be added. Although it's not a best practice to provision site columns or content types to a subweb, there's one key use case where it's very important to be able to do so - creating a lookup site column referencing a list in the subweb. Open to suggestions / modifications if there's a better alternative to this approach.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/FieldAndContentTypeExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/FieldAndContentTypeExtensions.cs
@@ -998,9 +998,12 @@ namespace Microsoft.SharePoint.Client
             var flink = contentType.FieldLinks.FirstOrDefault(fld => fld.Id == field.Id);
             if (flink == null)
             {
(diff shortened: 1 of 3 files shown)


PaoloPia:
Hi @TheJeffer,
We discussed internally (in PnP Core Team) and currently, we would like to keep this option out, because from a guidance perspective and from a best practices perspective it is not good to have site columns in subwebs, as well as it is not good to have lookup columns targeting lists in subwebs. Thus, I'm sorry but we can't merge this PR. Nevertheless, thanks for your efforts, which we truly appreciate.

PR #1381, opened 2017-10-04
Communication and Team Site Collection Creation

erwinvanhunen:


gautamdsheth:
@erwinvanhunen  - thanks a lot for this 👍 . However, you are passing the metadata type as `SP.Publishing.PublishingSiteCreationRequest`, whereas according to [this](https://docs.microsoft.com/en-us/sharepoint/dev/apis/communication-site-creation-rest), it should be `SP.Publishing.CommunicationSiteCreationRequest`.  So, what is the correct one ? I tried using `CommunicationSiteCreationRequest` and it gives error but if i try `PublishingSiteCreationRequest` it works. A bit confused here ! Is the API going to change in future ?

PR #1335, opened 2017-09-12
Support list by path and title

wobba:
| Q               | A
| --------------- | ---
| New feature?    | yes
| Related issues?  | Includes #1333 

#### What's in this Pull Request?

* Support token replacement in JsonControlData of client side web parts
* Support adding List web part based on list url only - for dynamic scenarios

Based the processing pattern on how xsltlistview wp are being processed.

@jansenbe let me know if you have any questions.

PR #1394, opened 2017-10-11
Improve/fix list loading in List web part for client side pages

wobba:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| Related issues?  | #1335 

Changed to use extension method for loading the list, as a template export does not include the relative path.

Changed loading with id, to also load the list.


jansenbe:
Thx Mikael!

PR #1398, opened 2017-10-12
Fix to preserve attributes on view fieldrefs

TheJeffer:
Fixed to set ListViewXml to the inner xml of our view schema so that any additional attributes on FieldRefs that aren't directly settable through CSOM are properly set.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Currently, if you have a view fieldref like the following, the attributes aren't preserved on the view when provisioned:
`
<FieldRef Name="Role" listItemMenu="TRUE" linkToItem="TRUE" />
`

The easy fix for this is to set the ListViewXml property of the view to the inner XML of the view schema right after creating it. This will properly set the above attributes, allowing the user to specify that a field other than Title should be used as a link to open the list item and/or to display the menu for the list item.


PaoloPia:
Hi @TheJeffer,
This submission looks great! Thanks for your time and contribution. I've just merged it.

PR #1409, opened 2017-10-17
Fix to align extension method behavior with the SharePoint UI default behavior

PaoloPia:
By default, the SharePoint Online UI creates a Public Office 365 Group. The extension method should do the same thing, for consistency, IMHO.


erwinvanhunen:
good ones!

PR #1420, opened 2017-10-23
Added support for Site Classification

PaoloPia:
Added full support for Site Classification (activate, update, deactivate, get/set at site collection level)


stevebeauge:
This PR breaks compilation on On premise version of the engine. I guess it's missing one `#if`

jansenbe:
Fixed...it also broke our nightly test run...

PR #1438, opened 2017-11-06
Fixing data rows key column and overwrite behaviour

RodCoder:
| Q               | A
| --------------- | ---
| Bug fix?        | X
| Related issues?  | 
https://github.com/SharePoint/PnP-Provisioning-Schema/issues/259
https://github.com/SharePoint/PnP-Provisioning-Schema/issues/236

This pull request contains a fix for the data rows xml element. When defining the key column and update behaviour a new list item is always created.
The problem is that we're looking at the wrong key when iterating through each data row and verifying if dataRow.Key is null (which will always be null). In order to get the right value we need to look at the values inside the dataRow.Values and get the value where key == listInstance.DataRows.KeyColumn.

Example:
If we're updating items based on the column 'Title', we will have to iterate through each data row, get its values and grab the key/value pair correspondent to 'Title'. Once we have this pair, we can then build the Caml query using the value to get items where 'Title' matches the rows we're going to insert.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs
@@ -94,7 +94,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                                         // if it is empty, skip the check
                                         if (dataRowValues.Any())
                                         {
-                                            var query = $@"<View><Query><Where><Eq><FieldRef Name=""{parsedKeyColumn}""/><Value Type=""{keyColumnType}"">{dataRowValue.FirstOrDefault().Value}</Value></Eq></Where></Query><RowLimit>1</RowLimit></View>";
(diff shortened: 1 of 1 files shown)


RodCoder:
Thank you so much for this Patrick, it was basically just missing the "s". I've tested the solution in another project and only copied the lines for this specific fix, using VS Code and it didn't complain :)

patrick-blanc:
thx as well for the fix 😉

RodCoder:
Absolutely true, have to start doing these things after working hours :)

patrick-blanc:
@jansenbe  please take this pull request into consideration for next PnP PowerShell update

PaoloPia:
Hi @RodCoder,
Thanks for your contribution. Actually this fix doesn't fix the problem, I've just tested it right now and I still have duplicated rows. Nevertheless, you made us aware of a bug, which we will fix. Thanks for your time.

RodCoder:
@PaoloPia Let me have a proper look at it tonight. I've implemented this in my company and it doesn't duplicate rows, been using it for a month now. 
I've tried to extract only that fix but something can be wrong. If in any case you fix it first then better, the problem is identified :)

Issue #1457, opened 2017-11-17
The provisioning engine does not export list specific content type


stevebeauge:
#### Category
[X] Bug
[ ] Enhancement

#### Environment
[ ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[X] SharePoint 2013

If SharePoint on-premises, what's exact CU version: 

#### Expected or Desired Behavior
The engine should support list local content types customization. Especially, fields created locally in a list and added to a content type should be exported.

#### Observed Behavior
List specific content types are ignored. Probably, this a missing feature of the provisioning schema.

#### Steps to Reproduce
1. Create a site content type `My content type`, inheriting `Element`
1. Create a list `MyList` based on a generic list
1. Enable content types on the list
1. Add `My content type` to the list
1. Create a field `MyField` **in the list** (no site column). Uncheck `Add to all content types`
1. Edit the **list** content type `My content type` (not the site content type)
1. Add the field `MyField` to the content type
        > At this point, the list settings show the field `MyField` as referenced in `MyContentType`
1. Export the provisioning template (Fields,ContentTypes and Lists handlers are enough)
        > there's no mention of `MyField` referenced to `MyContentType`

ruslanurban:
Also, there are cases when you need to display a different label in the "New Document" menu. Creating a list-level content type would facilitate this requirement.

Crispify:
Any work on this done already? I'm working with BDC fields (External Data) that cannot live on site collection level, so I'm adding it on library level and adding it to some of the content types on library level. The schema currently has no way of specifying this situation and when cloning the site, it leaves the BDC field disconnected of all content types.

jonathaneckman:
+1 but for SharePoint Online.

hasannaqvi83:
Any update on this ?

I recently have noticed that PE doesn't facilitate adding fields to content types local to a list.

stevebeauge:
Unfortunately, nothing has been done for that (as far as I known).

I guess the issue won't be solved, because it would require fixing the schema first then perform a deep rewrite of core handlers. The team behind the project is clearly under effective to handle the increasing number of issues and submitted PR. I wouldn't bet on a resolution.

userings:
+1 for SharePoint 2016

PaoloPia:
Hi,
[...]

Issue #1353, opened 2017-09-21
A field added to a list shouldn't be added to content types 


stevebeauge:
#### Category
[X] Bug
[ ] Enhancement

#### Environment
[ ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[X] SharePoint 2013

#### Expected or Desired Behavior

A field can be added to a list, without adding it to all content types (there's a checkbox for that in the SP UI).

When provisioning local fields using the Pnp engine, the fields shouldn't be added to the content types unless it was specified in the site used when exporting.

#### Observed Behavior

Local fields are always added to all content types in the list, even if the content type does not refers the field.

#### Steps to Reproduce

In the source site:

1. Create a list "some list" based on a generic list
1. Create a site content type "Some content type"
1. Enable CT management in the list
1. Add the "Some content type" CT to the list
1. Delete the Element content type
1. Add a local field to the list "Not in content type". Ensuse the checkbox "Add to all content types" is **not** checked
1. Export the site as template, using `Get-PnPProvisioningTemplate -Out c:\temp\pnp.xml
-handlers fields,lists,contenttypes`

At this point, the template should contains (I removed other artefacts):

````
<?xml version="1.0"?>
<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2017/05/ProvisioningSchema">
  <pnp:Preferences Generator="OfficeDevPnP.Core, Version=2.19.1710.0, Culture=neutral, PublicKeyToken=null" />
  <pnp:Templates ID="CONTAINER-TEMPLATE-BCE23B0BC20F49759271D6AD107D1B40">
    <pnp:ProvisioningTemplate ID="TEMPLATE-BCE23B0BC20F49759271D6AD107D1B40" Version="1" BaseSiteTemplate="STS#0" Scope="RootSite">
      <pnp:ContentTypes>
        <pnp:ContentType ID="0x010026D1CAA1F4DF694C89BE46D339AE18B8" Name="Some content type" Description="" Group="Debug" NewFormUrl="" EditFormUrl="" DisplayFormUrl="">
          <pnp:FieldRefs>
            <pnp:FieldRef ID="c042a256-787d-4a6f-8a8a-cf6ab767f12d" Name="ContentType" />
            <pnp:FieldRef ID="fa564e0f-0c70-4ab9-b863-0177e6ddd247" Name="Title" Required="true" />        
          </pnp:FieldRefs>
        </pnp:ContentType>
[...]

PR #1458, opened 2017-11-17
Partial fix of issue #1353 related to local fields

stevebeauge:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no 
| New sample?      | no 
| Related issues?  | partially #1353 and #1457 

This PR partially fixes local field regression.

Here is the status of what's working or not (in scenarios where PnP engine is use to "clone" site, by exporting the template and applying it):

[X] List with content types not enabled are cloned properly (site fields and local fields)
[ ] List with content types enabled : 
    [X] site fields referenced in the site content types are cloned properly
    [ ] all list local fields are cloned, but are not bound to content types
    [ ] all site fields are cloned, but are not bound to content types if the reference is in the list content type

I choose to still submit the PR as it solve the first case below, as it is widely used. Unfortunately, the other cases can't be fixed unless the schema is also extended.


jansenbe:
Thx @stevebeauge

PR #1469, opened 2017-11-23
Added support to get an app by its title

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | no 
| New feature?    | yes
| New sample?      | no
| Related issues?  | NA

Currently you can either get all available apps or get an app by its ID. (due to API "limitation")
So, have kinda "added support" to get an app by its title.

In the future, if and when we get API support to get an app by title, we can simply replace some code in the BaseGetAvailableAsync method without impacting the implementation outside.

PS - Would be great if we also have API support (some time in the future)  to get an app by its title, something along the lines of 
`/_api/web/tenantappcatalog/AvailableApps/GetByTitle`


gautamdsheth:
Closing this for now. Will be adding it again as I need to update my fork

PR #1471, opened 2017-11-27
Using TokenParser on CustomAction.RegistrationId

luismanez:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

This PR includes code to use the TokenParser on CustomAction.RegistrationId, as it makes sense to resolve tokens there (as List ID, specially useful for registering spfx extensions, as RegistrationType=ContenType is not yet supported, so we need to specify a ListID if we want to use the extension with a specific List).


PaoloPia:
Hi @luismanez,
Thanks for your contribution. It has been merged!

PR #1476, opened 2017-11-29
Bugfix: infinite loop when applying a template with a workflowdefinition that already exists

okms:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| New sample?      | no
| Related issues?  | --

fixing issue causing infinite loop when applying a template with a workflowdefinition that already exists, typically in an update scenario. Also adding support for updating workflowdefinitions

The infinite loop issue was intruduced with change dd917b3ea78890e29ea6fe6a477007f94bf341b6 as far as I can tell. On L247 the break is inside the if clause, so an exisiting workflow definition will cause the loop to never end.

cc: @PaoloPia 

#### Guidance
*You can delete this section when you are submitting the pull request.* 
* *Please update this PR information accordingly. We'll use this as part of our release notes in monthly communications.*
* *Please target your PR to Dev branch.*


PaoloPia:
Thanks @okms I will double-check it and merge it.

PaoloPia:
Thanks @okms I included your fix with some others, all related to the workflow provisioning. I will close this item, but you will be credited for your contribution. Thanks again.

PR #1490, opened 2017-12-06
Avoid duplicate image renditions if provisioned multiple times

gbouveret:
…me name, width & height)

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Define a template with an Image Rendition, and provision it twice for the same site : the rendition will be present 2 times. This is because of the "Contains" test that is not relevant for this kind of object. I'm using a LINQ query to test if there is an existing image rendition (same name, width & height) in order to avoid provisionning multiple times the same image rendition.


jansenbe:
Hi @gbouveret, nice catch. Will merge this one so it get's included in the upcoming December version.

PR #1505, opened 2017-12-21
Optimized default value for folder code

### LINKED ISSUE #1494: SetDefaultColumnValues Performance Degredation on Large Libraries
#### Category
[X] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ ] SharePoint 2013

Performance should be the same on the first or 800th term. 
#### Observed Behavior
The code in the SetDefaultColumnValues method is using enumeration over the LocationBasedDefaults file and as the file grows performance degrades. After about 500 items the code begins to timeout.

#### Steps to Reproduce
I have a trial tenant that I can provide access to, but the steps are:

1. Create a library with 1000 folders.
2. Create an MMS term for each folder.
3. Loop through the MMS Terms and apply the Default value for each term to each folder using Set-PnPDefaultColumnValues in PowerShell

wobba:
| Q               | A
| --------------- | ---
| Enhancement/Bug fix?        | yes
| Related issues?  | fixes #1494 

When having many folders using the same field, the code loaded up the field for every single folder when parsing the existing settings. By adding a field name to field cache this made the code load one field one time only.

Also improved code to load the /Forms folder, not loading all folders, but access it directly. Reduces load time a lot when you have many folder in a library.

* Improved loading time of client_LocationBasedDefaults.html
* Added caching of Field
* Reduce loading terms from term store

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/ListExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/ListExtensions.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Text;
 using System.Xml;
 using System.Xml.Linq;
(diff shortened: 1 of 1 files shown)


jansenbe:
Thanks for the optimization Mikael.

PR #1506, opened 2017-12-21
Bugfix: merged terms may cause duplicate keys in termstore

okms:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

## Re-deploying merged terms causes the term store to become corrupt
I'm currently working on a project where we use PnP Provisioning Templates extensively. As part of the template for certain sites we deploy **Term Groups**. The guidance has been that it's fine to modify any of the shared terms that we deploy and we sometimes add new terms. The current implementation of the PnP engine is that any existing terms are left alone, and only new terms are provisioned.

This has been working fine for a couple of years, but this fall we started experiencing some issues when browsing the term store using the default term store admin UI, but also when redeploying

 Taxonomy ui error

Error showing up in PowerShell

This error was both reproducible on our internal onprem environment and in Office 365. The common denominator was that the same template had been deployed both places, and manual modifications had been done to the TermSet between deployments.

It took a while to figure out, but in the end we found that merging two terms which are deployed deployed with PnP/Created with CSOM will cause this error the next time the term is redeployed. I.e the steps to reproduce are as follows

* Deploy a template containing two or more terms
* Merge two of the terms originating from the template
* Re-deploy template 

Tracing through what happens internally when provisioning term groups, I found that PnP requests all child tems from any parent term and checks if what we have in our template exists in the term store. Merged terms are of course missing from the list of terms, as they are hidden more or less entirely from end users and scripts alike. We will find a reference to the missing Id under the destination terms _MergedTermIds_ property, and we can also see that a term has been merged if we get another term in return when using TermStore.GetTerm(_guid_). 

Knowing this I've added a just in time check when creating a new term where we do a request to the term store to see if we get another term in return or if indeed the term does not exist at all.

I've attached a [test case](https://github.com/SharePoint/PnP-Sites-Core/files/1579301/tax-testcase.zip) you can run using PnP-PowerShell. The procedure is as follows

* Connect to  a site

  ```
  Connect-PnPOnline -Url https://tenant.sharepoint.com/
  ```

* Apply the provisioningtemplate attached

  ```
  Apply-PnPProvisioningTemplate -Path .\template.xml -Handlers TermGroups
  ```

* verify that you have  a new term group in you termstore called _TermGroup Foo_ with the term set _Termset Foo_ 

* Run the script `Merge-Terms.ps` , the script assumes you are still connected to your site
  **OR**
  Use the UI to merge two terms, e.g. Foo 1 and Foo 2

* Re-apply the provisioning template  
  ```
  Apply-PnPProvisioningTemplate -Path .\template.xml -Handlers TermGroups
  ```

You will now observe that the Term **Term Foo** has a duplicate child element. You may also observe that you get an error message when trying to expand the child elements of this term (may require that you refresh the page one or more times)


okms:
@VesaJuvonen had any chance to look into mergin this PR?

okms:
Any updates?

wobba:
Thank you @okms!

PR #1515, opened 2018-01-02
Fix missing call to token parser when querying for key column value.

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix missing call to token parser when querying for key column value.


jansenbe:
Thanks for this improvement Jens. It makes sense to also parse the key value.

PR #1558, opened 2018-01-30
Add site title token and proper implementation of site name token

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | (yes)
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

Site name token is incorrectly implemented and returns title. This has been fixed by adding a new site title token that returns the site title. The site name token has been modified to return the site name.
An extension method has been made available on the web to get the site name.

This change will unfortunately change value that the site name token used to return. It's unfortunate that this token was introduced with this oversight in the first place. I see no good naming scheme for the correct implementation of site name token and simultaneously retain backwards compatibility.

I leave it up to the PnP team to decide what approach to go for. Feel free to rename the token value as required. The important thing is to have a site name token and a site title token that returns expected values.

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Extensions/WebExtensionsTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Extensions/WebExtensionsTests.cs
@@ -11,6 +11,7 @@ using System.IO;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
 using OfficeDevPnP.Core;
 using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers;
+using OfficeDevPnP.Core.Entities;
 
 namespace Microsoft.SharePoint.Client.Tests
 {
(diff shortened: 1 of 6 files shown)


erwinvanhunen:
Hi!

While I value the approach you took, we will not be able to merge this one, as it changes the behavior (as you stated) of the SiteName token. This would mean that potentially thousands of existing templates 'out there' would be invalidated in their behavior. Unfortunately that is not acceptable from backwards compatibility reasons.

In your implementation you return a Url for the sitename token. I fail to understand how that is more correct than what it was?

How do your changes differ from the {site} and {sitecollection} tokens?

For now I will close this PR. If you feel that your PR still stands, please do not hesitate to reopen it.

jensotto:
Site name is defined as the last path component in the URL. This is not the same as the {site} or {sitecollection} tokens as they return the complete server relative URL of the site.
Example:
given a site with server relative URL "\sites\site1\subsite1" and title "Sub Site 1"
{site} returns \sites\site1\subsite1
{sitecollection} returns \sites\site1
{sitename} SHOULD return subsite1 but incorrectly returns "Sub Site 1". Microsoft has implemented SPWeb.Name to return subsite1 and SPWeb.Title to return the web title. This should be the same for the tokens.
{sitetitle} is therefore correct token to return the site title.

I said it was up to you to find a way to avoid the backwards compatibility issues. I do not want to take that decision as I feel the current token naming is incorrect and confusing.

jensotto:
How do I re-open this PR?

erwinvanhunen:
Notice that this PR needs changes in the token implementation in order for us to merge it. We cannot accept breaking changes to the tokens (breaking change in this case is that if a token returns a different type of value from what it used to do), regardless if the current implementation is wrong or right. If the current implementation is wrong is debatable in my opinion. The tokens in the provisioning engine do not necessarily represent values of properties in the CSOM object model. But we can agree to disagree there. If you think that the current implementation of the SiteName token is wrong you will have to create a new token instead to implement your functionality. It's not enough to introduce a new token with the old functionality of the current SiteName implementation with a new name.

jensotto:
I totally understand, but I didn't want to make the decision for what token to use. However I have made a change that possibly could work. I still have moved the logic into another token class to properly name it internally at least. I've added new token value for site title so that at least that can be used as expected together with the old sitename token value. I've renamed the token value for SiteNameToken to webname. 
I hope this solves the backward compatibility issues.

erwinvanhunen:
Thank you. It has been merged manually, I made some minor changes. Please check out my PR to see what I changed.

jensotto:
Thank you. I totally forgot the TokenParser update

PR #1582, opened 2018-02-13
Fix TaxonomyExtensions to work with other languages than English

stevebeauge:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no 
| Related issues?  |

This PR fix taxonomy helper when the store has not english language defined. Instead, the code should use the default store's language.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs
@@ -519,7 +519,7 @@ namespace Microsoft.SharePoint.Client
             TermStore ts = tSession.GetDefaultSiteCollectionTermStore();
             TermSet tset = ts.GetTermSet(termSetId);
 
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Hey @stevebeauge, nice change , how about also modifying it in the `AddTermToTermset` method [here](https://github.com/SharePoint/PnP-Sites-Core/blob/bcd9d7582e26f7937518ba58852b6fb861caa23c/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs#L522) as well.

Can you modify that line to as well ?

`t = tset.CreateTerm(term, ts.EnsureProperty(tstore=>tstore.DefaultLanguage), termId);`

stevebeauge:
@gautamdsheth : I've added the code as you suggested.
However, there's still occurrences of 1033 in the code. I'm not sure how to patch it so I'll leave as it.

PaoloPia:
Thanks @stevebeauge, great work.

PR #1584, opened 2018-02-14
Current user can't be removed from new SecurableObject role assignments

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

If a new SecurableObject (list, list data row, file or page) is created and the role inheritance is broken it's impossible to remove the current user from the role assignments. This happens because the securable objects role assignments are fetched before the securable objects role assignments inheritance is broken. In CSOM the current user is always added if copy role assignments are set to false.
For some reason this issue is only present in SharePoint 2013 (maybe 2016 too?) but not in O365.

Regardless it's a bug that the role assignments are fetched before the role assignments inheritance is broken.

Also fixed a using statement that is not valid On-Prem.

This PR is somewhat related to #1529 Please see comment on line 156 in ObjectPagesTest.cs


jensotto:
I originally incorrectly thought that this only happened for pages, but I have confirmed that this happens for all SecurableObjects except for Web as this is already correctly implemented in ObjectSiteSecurity.
The impact of this bug is therefore bigger that first assumed.

jensotto:
I've changed the test to use the role definition token.
@VesaJuvonen is there any chance of having this PR included in the upcoming March release?

jensotto:
@VesaJuvonen Is it possible to include this fix soon?

PR #1587, opened 2018-02-14
Fix Event item provisioning

### LINKED ISSUE #1586: Apply-PnPProvisioningTemplate fails when provisioning an Event item

*Please navigate to the appropriate repository by clicking on the link below and create your issue there. You can delete this section after you have navigated to the correct repository.*

#### Category
[x ] Bug
[ ] Enhancement

#### Environment
[ x] Office 365 / SharePoint Online
[x ] SharePoint 2016
[ x] SharePoint 2013

PnP Template successfully creates an Event item in a Calendar list.

#### Observed Behavior

I receive the following Error:

```
Invalid data has been used to update the list item. The field you are trying to update may be read only.
```

In order to create an event item the ```EndDate``` and ```EventDate``` must to be set at the same time. Looking at the ```ObjectListInstanceDataRows``` code the ```listItem.Update()``` is being called per field, rather than once per item, causing the above exception.

#### Steps to Reproduce

Using SharePointPnPPowerShellOnline module version 2.23.1802.0, run Apply-PnPProvisioningTemplate using the following template:

```
<?xml version="1.0" encoding="utf-8" ?>
<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2018/01/ProvisioningSchema">
  <pnp:Preferences>
  </pnp:Preferences>
	<pnp:Templates>
    <pnp:ProvisioningTemplate ID="CreateEventItemTest">
      <pnp:Lists>
        <pnp:ListInstance Title="Calendar"
            Description=""
            DocumentTemplate=""
            TemplateType="106"
            Url="Lists/Calendar"
            EnableModeration="true"
            MinorVersionLimit="0"
            MaxVersionLimit="0"
            DraftVersionVisibility="2"
            TemplateFeatureID="00bfea71-ec85-4903-972d-ebe475780106"
            ContentTypesEnabled="true"
            EnableFolderCreation="false">
          <pnp:DataRows>
            <pnp:DataRow>
              <pnp:DataValue FieldName="Title">Test Event</pnp:DataValue>
			  <pnp:DataValue FieldName="EventDate">2018-02-13 15:00:00.000</pnp:DataValue>
			  <pnp:DataValue FieldName="EndDate">2018-02-13 15:30:00.000</pnp:DataValue>
            </pnp:DataRow>
          </pnp:DataRows>
        </pnp:ListInstance>
      </pnp:Lists>
	  </pnp:ProvisioningTemplate>
	</pnp:Templates>
</pnp:Provisioning>
```

craig-blowfield:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #1586 

Calling listItem.update() per field value causes an exception when creating an Event item via a template. Moving this out and calling it once per item resolves this issue, although this could have side effects.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs
@@ -246,9 +246,10 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                                                         listitem[parser.ParseString(dataValue.Key)] = fieldValue;
                                                         break;
                                                 }
-                                                listitem.Update();
                                             }
                                         }
+
+                                        listitem.Update();
                                         web.Context.ExecuteQueryRetry(); // TODO: Run in batches?
 
                                         if (dataRow.Security != null && (dataRow.Security.ClearSubscopes == true || dataRow.Security.CopyRoleAssignments == true || dataRow.Security.RoleAssignments.Count > 0))


craig-blowfield:
@jensotto I don't see too many benefits in partially creating a list item when subsequently the template fails anyway. I would rather be able to create a list item or not at all

Were the reasons for having an update per field to partially create list items or another reason, such as:
* Possibly more granular errors, per field value?
* Cross-field value usage?

jensotto:
@craig-blowfield I don't know the reason behind the current implementation. I'm just pointing out that for this to work there can be no calls to ExecuteQuery inside the foreach loop. This shouldn't be too complicated as the operations that I can see (calls to web.EnsureUser()) could easily have been retrieved before entering the foreach loop. This way it will be possible to call list.Update only once.
If you could make this change too, then I guess your suggestion would work as expected.

jensotto:
@PaoloPia are you sure that this will work? Did you read my comments? I am afraid that this change will break list instance datarow provisioning when there are user fields involved.

PaoloPia:
@jensotto what you are referring to is a more general issue, not only with User fields and we're working to fix it right now.

jensotto:
@PaoloPia I'm not sure what you refer to. The only thing I can remember I thught about when initially replying to this PR was that all code inside the for each loop would be checked for usage of ExecuteQuery (possibly in any extension methods). I however couldn't see any other usages of ExecuteQuery than the two inside the Users field type section.

jensotto:
@PaoloPia sorry, just disregard my last post. I misread your post and thought that you where asking me for clarification of something :)

PaoloPia:
It is now fully fixed.

jensotto:
@PaoloPia Yes, I just saw the recent commit. I'm afraid that it has some issues. I'm not sure if there are relevant tests in place for testing list instance data rows and taxonomy fields, but as far as I can see the taxonomyfield modifications done in the commit will not work properly.

PaoloPia:
We tested the solution with single and multi-value taxonomy fields, as well as with event items, and with multiple User fields on a single item. Everything was ok. Can you be more specific about the issues that you foresee in the code? Thanks.

jensotto:
Mainly I believe it's the situation where you specify a blank value (clearing out a preexisting taxonomy field value). For that to be successful you need to do the workaround that I have done in TaxonomyExtensions.SetFieldValueByLabelGuidPair. The collection taxonomyValues will also be empty and you will get an exception on line 279.
There will also be problems if you set a TaxonomyFieldTypeMulti first and then a TaxonomyFieldType or any other field that calls ExecuteQuery().

Also I see now that I have overseen the consequence of calling ExecuteQueryRetry() in my contribution in SetFieldValueByLabelGuidPair. My intention was to make it possible to set multiple field values and let the user call item.Update() and ExecuteQuery when all field values has been set. This is how the SSOM method works. By calling ExecuteQueryRetry I effectively end up in the same situation that you now try to fix. A actually do not know if it's possible to get around unless Microsoft fixes the methods exposed on TaxonomyField for setting the field values (SetFieldValueByValueCollection and SetFieldValueByValue) and the TaxonomyFieldValueCollection and TaxonomyFieldValue. I'll have a look but I guess you will try to have a release tomorrow, and that might be to short time for me to try to find a solution.

erwinvanhunen:
Wouldn't line 144 catch the request to clear out values?

jensotto:
No, you also need to set the taxonomy hidden field. It's also necessary to trigger the taxonomy field list item event receiver and let it execute properly. This can only be done by executing SetFieldValueByValueCollection or SetFieldValueByValue. The only way I've found to do this when clearing the taxonomy field value is to use the workaround that I have in SetFieldValueByLabelGuidPair. I spent many hours to debug and understand the code flow in the event receivers and SetFieldValueByValueCollection and SetFieldValueByValue methods to find an approach that works.

jensotto:
One additional note is that when testing taxonomy field value modifications it's important to check that the hidden field is updated as expected and that the TaxCatchAll field has correct values. TaxCatchAll can only be verified it there are at modifications to at least two different taxonomy fields. If TAXCatchAll is not updated properly you will have issues with search, and probably more.

jensotto:
Please have a look at the test I made for this: TaxonomyExtensionsTests.SetBlankTaxonomyFieldValueTest

jensotto:
The reason you need to modify at least two taxonomy fields is that the taxonomy field list event receiver will fail if taxonomy fields are not updated properly. When this happens the event receiver will just stop updating the TaxCatchAll field and will often then only have one entry. It's therefore necessary to modify two taxonomy fields to ensure that this situation does not happen.

jensotto:
@erwinvanhunen @PaoloPia  I think I have a solution for the SetFieldValueByLabelGuidPair method. I just want to run a few more tests. The method will not have any calls to ExecuteQuery but I need to duplicate the code from 
Microsoft.SharePoint.Taxonomy.TaxonomfyFieldValue.PopulateFromLabelGuidPair

If it's possible to change the data row function to execute SetFieldValueByLabelGuidPair in the new foreach block that you created at line 299 and store the raw data in lines 279 and 283 instead of using TaxonomyFieldValueCollection I think we are in the clear.

erwinvanhunen:
I was looking at a similar approach, by just storing the old fashioned taxonomy string -1;#{Label}|{guid} in an array instead and then building the things later. I'm open to suggestions for your end, so keep us in the loop

erwinvanhunen:
I got provisioning to work, regardless of order of the fields in the datarows. Haven't looked yet at clearing existing values. Up next now.

erwinvanhunen:
have a look here: https://github.com/erwinvanhunen/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs. I'm trying to optimize it a bit more, but this code works over here. Including cleaning existing taxonomy field values

jensotto:
I think there will be issues when using #; and the term label guid pair. The problem lies in that the WSS id that normally is before the #; token is missing, and SharePoint will parse that to 0. When this value is anything but -1 the WssID will not be correctly set and the link to TaxonomyHiddenList will not be set up and a new taxonomy value will not be added to that list either. It could possibly be remedied with adding -1 before #;
Will this approach not partly defeat the original PR made by @craig-blowfield ? There will be multiple list item updates.

erwinvanhunen:
I think right now it's more important to have it working than to make it optimal. Given the deadline we're working against.

Notice that I hardcode the -1 WssId in the string, so it's not missing.  Line 275 (https://github.com/erwinvanhunen/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs#L275)

erwinvanhunen:
regarding multiple list item updates, sure, not nice. The engine in general is already pretty chatty in general, and in this case we only introduce an extra roundtrips if you set taxonomy fields.

jensotto:
I agree. It looks OK now except that I'm worried that when a blank value is set in the single value taxonomy field the update will not be successful. Have you checked that the hidden field (TextField) and TaxCatchAll field are properly updated?

jensotto:
If you use the code block that I have here https://github.com/erwinvanhunen/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs#L1694-L1715 when the term label and term guid is empty then I think it will work when setting a blank value.

erwinvanhunen:
Yes, the hidden fields and taxcatchall fields get updated correctly

jensotto:
Wouldn't it have been possible to use the ClientContextExtensions.Clone method to create a new context and run a separate set of queries against SharePoint?

jensotto:
Have you checked with more than one taxonomy field in the list item? The problem only happens if at least two taxonomy fields have values and one of them is set to empty/cleared

erwinvanhunen:
yes I have. I will merge this code now.

jensotto:
@erwinvanhunen I've created a test method to test this scenario and I get an exception when setting a single value taxonomy field to a empty string. This is because of the code in line 337 https://github.com/erwinvanhunen/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstanceDataRows.cs#L337 There will not be two items in the array if the string is empty.
I also tried using a string with only the pipe character "|". This leads to a server exception. The code does not work when setting a blank value and the only workaround is to use the code I mentioned earlier: https://github.com/erwinvanhunen/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs#L1694-L1715
It's probably too late to fix this before the release now, but I'll see if I can improve this for the next release.

jensotto:
The test is available here https://github.com/jensotto/PnP-Sites-Core/blob/list-data-row-verifications/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectListInstanceTests.cs#L190
It's currently a bit messy, but I didn't have time to clean it up.

jensotto:
If you have tested with setting the data row value to null, then that will also fail as the TaxCatchAll field will be cleared an loose the data from all the other taxonomy fields in the list. This is a bit critical, because it's not immediately obvious. It will lead to missing search results possibly other issues too.

erwinvanhunen:
Template snippet 1

```xml
<pnp:DataRows KeyColumn="ID" UpdateBehavior="Overwrite">            
  <pnp:DataRow>
    <pnp:DataValue FieldName="ID">25</pnp:DataValue>
    <pnp:DataValue FieldName="Title">test</pnp:DataValue>
    <pnp:DataValue FieldName="_x0074_2">Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd</pnp:DataValue>
    <pnp:DataValue FieldName="_x0074_1">Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd;Internal Test|3f3e02d3-93d4-4d8e-a6ef-d04fc9f71b73</pnp:DataValue>	      
  </pnp:DataRow>
</pnp:DataRows>
```

Values after applying:

```
_x0074_1                         {590cadf5-7ff2-4258-bf5c-d810ad41fbfd, 3f3e02d3-93d4-4d8e-a6ef-d04fc9f71b73}
c2c894a1aa2343d28cfbdd40d63da0bc Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd;Internal Test|3f3e02d3-93d4-4d8e-a6ef-d04fc9f71b73
TaxCatchAll                      {2, 1}
_x0074_2                         Microsoft.SharePoint.Client.Taxonomy.TaxonomyFieldValue
le5a6acca0ab4826821df06a0622660e Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd
```

Template Snippet 2:

```xml
<pnp:DataRows KeyColumn="ID" UpdateBehavior="Overwrite">            
    <pnp:DataRow>
      <pnp:DataValue FieldName="ID">25</pnp:DataValue>
      <pnp:DataValue FieldName="Title">test</pnp:DataValue>
      <pnp:DataValue FieldName="_x0074_2"></pnp:DataValue>
      <pnp:DataValue FieldName="_x0074_1">Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd;Internal Test|3f3e02d3-93d4-4d8e-a6ef-d04fc9f71b73</pnp:DataValue>
    </pnp:DataRow>
</pnp:DataRows>
```

Values after applying:

```
_x0074_1                         {590cadf5-7ff2-4258-bf5c-d810ad41fbfd, 3f3e02d3-93d4-4d8e-a6ef-d04fc9f71b73}
c2c894a1aa2343d28cfbdd40d63da0bc Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd;Internal Test|3f3e02d3-93d4-4d8e-a6ef-d04fc9f71b73
TaxCatchAll                      {2, 1}
_x0074_2
le5a6acca0ab4826821df06a0622660e
```

I now reapply Template Snippet 1. 

Then I apply snippet 3:
```xml
<pnp:DataRows KeyColumn="ID" UpdateBehavior="Overwrite">            
  <pnp:DataRow>
    <pnp:DataValue FieldName="ID">25</pnp:DataValue>
    <pnp:DataValue FieldName="Title">test</pnp:DataValue>
    <pnp:DataValue FieldName="_x0074_2">Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd</pnp:DataValue>
    <pnp:DataValue FieldName="_x0074_1"></pnp:DataValue>	      
  </pnp:DataRow>
</pnp:DataRows>
```

Values after applying:
```
_x0074_1                         {}
c2c894a1aa2343d28cfbdd40d63da0bc
TaxCatchAll                      {1}
_x0074_2                         Microsoft.SharePoint.Client.Taxonomy.TaxonomyFieldValue
le5a6acca0ab4826821df06a0622660e Internal|590cadf5-7ff2-4258-bf5c-d810ad41fbfd
```

erwinvanhunen:
I assume you have tested it too with the latest commit from last night. If you receive different results I would love to see those. Thanks! :-)

jensotto:
The results looks fine. I think the reason that it works is that you reapply the value to all taxonomy fields in the item. It's possible that the issue only happens if there are taxonomy fields in the list item that has value and the provisioning template only clears one of them. This is the scenario I have tested in the unit tests I provided for the extension methods. If that's the case it could be argued that the impact is less important as long as all taxonomy fields are modified when applying the provisioning template. It should definitively be fixed, but since the current version of PnP handles Taxonomy Fields as regular text fields it probably also suffers from this problem.

I did not use an XML template. I created the template in code. The result should be the same. Here are the result I got depending on the value of the data row field (single value taxonomy field):
"" -> Message: Test method OfficeDevPnP.Core.Tests.Framework.ObjectHandlers.ObjectListInstanceTests.CanClearTaxonomyFieldData threw exception: 
System.IndexOutOfRangeException: Indeksen var utenfor matrisegrensen.

"|" -> Message: Test method OfficeDevPnP.Core.Tests.Framework.ObjectHandlers.ObjectListInstanceTests.CanClearTaxonomyFieldData threw exception: 
Microsoft.SharePoint.Client.ServerException: creatingField

null -> Message: Assert.AreEqual failed. Expected:<1>. Actual:<0>. TaxCatchAll does not have 1 entry

I used this test method: https://github.com/jensotto/PnP-Sites-Core/blob/list-data-row-verifications/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectListInstanceTests.cs#L190

erwinvanhunen:
Yes, you are correct. In the case where we only clear one field it empties the catchall field, regardless of other values being present in other fields.

erwinvanhunen:
There was another logic mistake in the code I merged yesterday. Working on fixing that.

erwinvanhunen:
https://github.com/SharePoint/PnP-Sites-Core/pull/1639/files

erwinvanhunen:
in my tests it clears out the tax field, while retaining the taxcatchall with this commit.

jensotto:
Yes, I think you nailed it now. There are however two edge cases that probably should be included in the test at line https://github.com/SharePoint/PnP-Sites-Core/pull/1639/files#diff-d77d8c1d496ecf5dd4afa444a2c591c7R341
Check for empty string and check for the pipe character "|". Both of these values should result in clearing the taxonomy field.

PR #1592, opened 2018-02-19
Fix group description update during site security provisioning

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix incorrect update of site group description during provisioning. 
There are two places that holds the description for a group:
1. The group object Description property
2. The Note field for the group entry in the User Information List (UIL)

SharePoint UI uses the content in the Note Field from UIL to render the group description text in the UI.
It's therefore critical that the UIL is always updated with any group description changes.

Previously only the site group description would be set if the description was plain text. The entry in UIL was not set.
If the group was created with a rich text description both the group description and the UIL entry would contain the rich text entry. This is a bug as the group description property should only contain plain text.
The previous implementation had a bug where if you tried to update the group description with a plain text description it would never do the update if the group had no description already set. This was due to an incorrect check for null or empty string on the group description instead of checking the template group description.
Finally the group description would only update UIL if a rich text description was supplied.

Implementation has changed so that UIL and group description properties are always held in sync. This is done by using the ConvertSimpleHtmlToText method in Microsoft.SharePoint.Utilities.SPHttpUtility.

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
@@ -146,6 +146,7 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
             string htmlGroupName = string.Format("Test_HTML_{0}", DateTime.Now.Ticks);
             string plainTextDescription = "Testing plain text";
             string richHtmlDescription = "Testing HTML with link to <a href=\"/\">Root Site</a>";
+            string richHtmlDescriptionPlainTextVersion = "Testing HTML with link to Root Site";
 
             var template = new ProvisioningTemplate();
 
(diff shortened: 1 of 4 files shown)


jensotto:
One additional observation is that it seems that SharePoint never updates the group description properties other than when creating new site collections. See Microsoft.SharePoint.Utilities.SPUtilityInternal.CreateDefaultSharePointGroups
(from Microsoft.SharePoint.ApplicationPages.CreateSitePage.BtnCreateSite.Click)
and Microsoft.SharePoint.ApplicationPages.EditGroup.DoOperation and Microsoft.SharePoint.ApplicationPages.CBaseNewGroup.UpdateAdditionalProperties)

PaoloPia:
Hi @jensotto,
I skimmed through the PR. Before merging it, I'd like to see if we can use a different approach to convert HTML to plain text, rather than having a custom method. I put a flag "investigation needed" on this PR and I will process it in the upcoming days. Thanks for your effort.

gautamdsheth:
@jensotto  - How about using simply using Regex to remove the html tags ? AFAIK, it would work just fine, might probably need some deeper investigation if there is some corner case.

Something like:

`String plaintextDesc = Regex.Replace(parsedGroupDescription, "<.*?>", string.Empty);`

jensotto:
@gautamdsheth @paolopia The method is technically not custom as it is identical to the SSOM implementation as noted in the comments.
Regex would be difficult I believe. There are probably too many corner cases. Still it would be a custom method. The method I have proposed comes from SharePoints own implementation so it should be well tried and tested. That's not true with any other approach unless a third party solution can be found.
I think the best approach would be for Microsoft to expose the method in the CSOM library. The next best thing I believe is my approach where the SSOM method is duplicated in PnP. If any other approach is chosen extensive testing would be required to ensure identical behavior as SSOM. I don't se the benefit of this as that would require a lot of work.

PaoloPia:
Hi @jensotto, 
This PR has just been merged. Thanks.

PR #1599, opened 2018-02-20
Fix Incorrect List Tokens Used in Subweb

TheJeffer:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fixes issue where an incorrect list token would be used when provisioning a template with a list to a subweb if the root site had an existing list with the same name. If the token {listid:Some Shared List Title} was used in a page's web parts for instance, it would fail to provision saying the list was not found, because the token parser would replace the token with ID of the existing, root site list instead of your newly created list in your subweb.

PR adds a new overload to TokenParser.AddToken that allows you to optionally remove existing tokens with the same type and name before adding the new token, and updates ObjectListInstance to use that new overload when adding the tokens for newly created lists.


TheJeffer:
Encountered a few errors locally after implementing this fix. Closing this until I track down the cause to make sure this fix didn't introduce a regression.

TheJeffer:
Errors were completely unrelated. This is good to go.

jansenbe:
Hey @TheJeffer ,

Thanks for this PR...we did not do a good job in processing this one timeline and given that so much has changed over the last year we're not sure if this PR is still needed. Therefore we've decided to close this PR for now. If you still can reproduce the issue on the latest release then please create a new PR with updated code so we can evaluate it again. Apologies for the extra work this might be causing.

PR #1645, opened 2018-03-04
Fix bug #1640 - Method implementation is not atomic

### LINKED ISSUE #1640: Method implementation is not atomic.
#### Category
[x ] Bug
[ ] Enhancement

#### Environment
[ x] Office 365 / SharePoint Online
[ x] SharePoint 2016
[ x] SharePoint 2013

https://github.com/sharepoint/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs#L1672

Code sets field on listitem, however never commits this change through update and ExecuteQueryRetry. While it is to be assumed that the user will do this later in the calling flow, this is an assumption which is not guaranteed, especially given that methods in PnP are atomic, if not otherwise stated.

TeodoraI:
| Q                    | A
| --------------- | ---
| Bug fix?           | yes
| New feature?  | no
| New sample?  | no
| Related issues?  | fixes #1640

Bug #1640 - Method implementation is not atomic. Added update and execute in the method SetFieldValueByLabelGuidPair. Removed additional updates in methods SetTaxonomyFieldValue & SetTaxonomyFieldValues which use SetFieldValueByLabelGuidPair.


TeodoraI:
Fix #1640 - Method implementation is not atomic and removed not needed updates of the item

wobba:
@TeodoraI If you remove execute query on those two methods, you will break update for people using those methods. I do agree that an executequery should be added to `SetFieldValueByLabelGuidPair` though if that is the method you are using directly. An optimization could be to add `clientContext.HasPendingRequest` in the two existing methods calling `SetFieldValueByLabelGuidPair`.

TeodoraI:
@wobba, I reviewed again the two methods that use `SetFieldValueByLabelGuidPair` . 

First one - `SetTaxonomyFieldValue` has get query followed by execute. Then there is a if-else in both cases calling `SetFieldValueByLabelGuidPair` method which updates and executes the query. So this will not break and works as expected.

Second one - `SetTaxonomyFieldValues` has again get query followed by execute. Then we have if-else statement. The if statement contains string operations and calling `SetFieldValueByLabelGuidPair` method with the value from the string operations. In the else statement an error is thrown.

`SetFieldValueByLabelGuidPair` is not used anywhere else.

In conclusion, I do not believe removal will break anything because the code clearly does not have any pending requests.

wobba:
True that `SetFieldValueByLabelGuidPair` is not used anywhere else, but it's a public method in a public class, so anyone out there might be using it. If it had been internal, no problem.

TeodoraI:
I removed the update and execute in `SetTaxonomyFieldValue`  and `SetTaxonomyFieldValues` which do not have any pending requests or updates. Is there a reason why it is required?

wobba:
My bad.. took a closer look, and your code looks correct - ensuring an update for all of them.

Issue #1617, opened 2018-02-26
Deadlocks everywhere in PnP Core!!


OliverZeiser:
#### Category
[X] Bug
[X] Enhancement

#### Environment
[X] Office 365 / SharePoint Online

#### Expected or Desired Behavior
PnP Core methods should work no matter if they are called from an MVC or Console or whatever application...

#### Observed Behavior
Many of the newer methods in PnP Core library do not work when using them from an application that has an UI Thread. So putting it the other way, it only works from a console application and that includes the unit tests....
Newer methods in PnP Core that use the Rest API (/_api/XXX) instead of csom are using PnPHttpProvider or a different HttpClient and make async requests using await httpClient.SendAsync(...
But higher up the stack, those are often beeing called within a Task.Run(() => ....
This can lead to deadlocks when using PnP Core with applications that have an UI thread.
You can read more about this here for example: https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

#### Steps to Reproduce
Call the new AvailbleClientSideComponents() method from an MVC Application for example. Try to make multiple Calls and test it within a sync and also an async MVC Action. You will see that some requests will never bring there response back and end up in a deadlock.

```
            var page = new ClientSidePage(ctx);
            var components = page.AvailableClientSideComponents();
```

#### Possible Fix

Use httpClient.SendAsync().ConfigureAwait(false) or use async all the way up the stack and let us call the async methods directly!

seriewe:
Same Issue https://github.com/SharePoint/PnP-Sites-Core/issues/1537

jansenbe:
@OliverZeiser , @seriewe , @biste5 : I've updated the dev version to (hopefully) handle these deadlocks via https://github.com/SharePoint/PnP-Sites-Core/commit/5dc3e51ebb66c4c55f6f059a2dae0de03a9891ed. This is based on the work described in https://blogs.msdn.microsoft.com/benwilli/2017/02/09/an-alternative-to-configureawaitfalse-everywhere. Combined with that I've added some additional Async method versions as well.

Please give the dev branch version a try and see if fixes the issue in your applications. Thanks for testing.

OliverZeiser:
@jansenbe It seems to work, but there is one major issue now. I think you forgot to add the `await new SynchronizationContextRemover();`

asumner:
It appears as though this issue was resolved, why is it still open? @jansenbe

jansenbe:
Closing as this is resolved

PR #1352, opened 2017-09-21
Fixed deadlock on extracting webhook subscription on lists.

TanmayDharmaraj:
| Q               | A
| --------------- | ---
| Bug fix?       | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | #1343 

#### What's in this Pull Request?
Provisioning template deadlocks when extracting web hook subscriptions for lists via an async API. 
Had to use Task.Run specifically because it is not possible to make the ExtractObjects method async as it would warrant a change in all classes implementing this interface.

PR #1649, opened 2018-03-06
adding ConfigureAwait(false) for ExecuteQueryAsync calls

biste5:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | could be related to #1617 

In libraries, in general, it is good practice to use ConfigureAwait(false) with async/await to detach the calling context and avoid deadlocks
Here the fixed implementation for ExecuteQueryRetry

In some way already mentioned in #1352 , and here
https://twitter.com/lsspdev/status/962780144707481602

I know that it's actually quite a lot of references, for a small change, but please review the change too ;)

Massimo

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
@@ -69,9 +69,9 @@ namespace Microsoft.SharePoint.Client
         /// <param name="retryCount">Number of times to retry the request</param>
         /// <param name="delay">Milliseconds to wait before retrying the request. The delay will be increased (doubled) every retry</param>
         /// <param name="userAgent">UserAgent string value to insert for this request. You can define this value in your app's config file using key="SharePointPnPUserAgent" value="PnPRocks"></param>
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi Massimo, good feedback and indeed this is something we need to fix. Currently doing some research and looking into https://blogs.msdn.microsoft.com/benwilli/2017/02/09/an-alternative-to-configureawaitfalse-everywhere/ as way to have an easier/better fix.

biste5:
Hi Bert,
cool. I wanted to add some more details in my PR, but it can get a bit complex and long to explain.
Very good info anyway in that article. I have some experience on this done with mobile app development, where there it is more serious as sometime we deal with UI threads, and so you need to know where to detach context (with ConfigureAwait false), and where not to.

Feel free to ask for more info, or to discuss more directly, or if you need help in discussing a possible approach to fix, and eventual best practice for the future.

Massimo

jansenbe:
Closing this one as there's a (hopefully) correct solution. See notes of https://github.com/SharePoint/PnP-Sites-Core/issues/1617 for more context. We can re-open this PR if the done changes are insufficient.

PR #1650, opened 2018-03-06
Adding ConfigureAwait for GetRequestDigest helper method to avoid dea…

### LINKED ISSUE #1537: OfficeDevPnP.Core.ALM.AppManager.GetAvailable() livelock
#### Category
[ X ] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ ] SharePoint 2013

#### Observed Behavior
We use the ApplyProvisioningTemplate Web Extension from our code which ends in a Livelock if we call it from an async action!

The Root Issue is inside _OfficeDevPnP.Core.ALM.AppManager_  method _GetAvailable()_ which is called inside TokenParser.AddAppPackagesTokens(...)  
```
private void AddAppPackagesTokens(Web web)
    {
                ...
                var appPackages = manager.GetAvailable();

                foreach (var app in appPackages)
                {
                    _tokens.Add(new AppPackageIdToken(web, app.Title, app.Id));
                }
```

Async calls, like _GetRequestDigest, SendAsync_ and so one inside the  BaseGetAvailableAsync lead to a live lock.

```
` public List<AppMetadata> GetAvailable()
        {
            return BaseGetAvailableAsync(Guid.Empty).GetAwaiter().GetResult();
        }`
```

#### Solution
We add the ConfigureAwait(false) to all the async calls like mentioned here: https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f 

Now the ApplyProvisioningTemplate method runs as expected also from Controller and not just from the webjob!

The Method ConfigureAwait(false)  should be used for all the async methods inside the library or the method signature should be async!

biste5:
…dlocks

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | Possibly fixes #1537

As previous PR #1649 in relation to avoiding deadlocks of async using ConfigureAwait, here the addition of that in GetRequestDigest when it calls PnPHttpProvider.SendAsync

I say possibly fixes #1537 because I didn't verify myself it does provide a fix for it, and it isn't easy to find where the deadlock is caused in the current call stack. We should probably apply ConfigureAwait throughout all await calls (where possible).

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
@@ -533,7 +533,7 @@ namespace Microsoft.SharePoint.Client
                     {
                         request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", accessToken);
                     }
-                    HttpResponseMessage response = await httpClient.SendAsync(request);
+                    HttpResponseMessage response = await httpClient.SendAsync(request).ConfigureAwait(false);
 
                     if (response.IsSuccessStatusCode)
                     {


jansenbe:
Closing this one as there's a (hopefully) correct solution. See notes of https://github.com/SharePoint/PnP-Sites-Core/issues/1617 for more context. We can re-open this PR if the done changes are insufficient.

Issue #1117, opened 2017-03-10
Apply provisioning template on modern sites breaks when template includes 'Style Library'


biste5:
#### Category
[ X ] Bug
[ ] Enhancement

#### Environment
[ X ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

#### Expected or Desired Behavior
I have a modern site just created, for which I extracted its provisioning template using PnP-PowerShell.
I want to reuse this template to apply to other modern sites

#### Observed Behavior
The extracted template contains the following XML portion

```xml
...
 <pnp:ListInstance Title="Style Library" Description="Use the style library to store style sheets, such as CSS or XSL files. The style sheets in this gallery can be used by this site or any of its subsites." DocumentTemplate="" TemplateType="101" Url="Style Library" EnableVersioning="true" MinorVersionLimit="0" MaxVersionLimit="500" DraftVersionVisibility="0" TemplateFeatureID="00bfea71-e717-4e80-aa17-d0c71b360101" EnableAttachments="false">
          <pnp:ContentTypeBindings>
            <pnp:ContentTypeBinding ContentTypeID="0x0101" Default="true" />
            <pnp:ContentTypeBinding ContentTypeID="0x0120" />
          </pnp:ContentTypeBindings>
          <pnp:Views>
            <View Name="{C18A8232-EF5D-450B-98B6-312F2A78F5C1}" DefaultView="TRUE" MobileView="TRUE" MobileDefaultView="TRUE" Type="HTML" DisplayName="All Documents" Url="/sites/TestModern/Style Library/Forms/AllItems.aspx" Level="1" BaseViewID="1" ContentTypeID="0x" ImageUrl="/_layouts/15/images/dlicon.png?rev=44">
              <Query>
                <OrderBy>
                  <FieldRef Name="FileLeafRef" />
                </OrderBy>
              </Query>
              <ViewFields>
                <FieldRef Name="DocIcon" />
                <FieldRef Name="LinkFilename" />
                <FieldRef Name="Modified" />
                <FieldRef Name="Editor" />
              </ViewFields>
              <RowLimit Paged="TRUE">30</RowLimit>
              <JSLink>clienttemplates.js</JSLink>
            </View>
          </pnp:Views>
          <pnp:FieldRefs>
            <pnp:FieldRef ID="ccc1037f-f65e-434a-868e-8c98af31fe29" Name="_ComplianceFlags" DisplayName="Label setting" />
[...]

PR #1653, opened 2018-03-07
added higher async methods for branding extensions

baywet:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | continuation of the effort started in !1559

I wanted to get early feedback before doing everything @jansenbe @SharePointRadi 
Basically the goal is to provide higher level async methods so concurrency can be managed consistently from end to end.
I did a few things:
- reviewed the approach providing executeretryasync so we don't have to do conditional build all the way up and it'll be easy to provide that for on prem the day on prem CSOM has an async execute query as well.
- provided higher level async methods in branding extensions and switched things so they own the logic.
- wrapped call in the sync methods to provide backward compatibility

#### Guidance
We can either merge that now and author subsequent PR's if you think that's going to be easier or I can keep pushing changes to this one if you give me a go on the approach.
Happy to discuss about it :)


jansenbe:
Thanks @baywet. Getting more async methods is something we clearly need, especially now that we've a .Net Standard 2.0 version of the library that can be used in for example mobile development. Today I did a structural change (https://github.com/SharePoint/PnP-Sites-Core/commit/5dc3e51ebb66c4c55f6f059a2dae0de03a9891ed) for all async methods, which would be something to apply to all methods if it receives positive test feedback from community (to avoid deadlocks). We'll internally discuss about the best approach to move this forward and will let you know asap.

baywet:
Thanks for getting back to me this quick.
Looked at the commit it makes sense. 
I still believe we should avoid conditional build for the ExecuteQueryRetryAsync. Provide it to everybody even if it's not doing async yet because CSOM on prem doesn't have the method.
It'll give people time to remap their code and benefit the performance improvement when we ship an update

baywet:
Hi @jansenbe what was the outcome of the tests you ran for async support? how can we move forward?

baywet:
Refering thed deadlocks and async issues as discussed on the call.
https://github.com/SharePoint/PnP-Sites-Core/issues/1617
https://github.com/SharePoint/PnP-Sites-Core/issues/1537
https://github.com/SharePoint/PnP-Sites-Core/issues/1117
Let me know if I missed anything @jansenbe

jansenbe:
Hi Vincent,

As discussed during the community call thanks for this PR. The current async calls have been tested and the model is working fine so far, hence I would like to ask you to adjust this PR (or create a new one) based upon this model:
- Add `await new SynchronizationContextRemover();` in each publicly called async method, before actual method logic but after argument validation
- Use `Task.Run(() => AddAsync(file, filename, overwrite)).GetAwaiter().GetResult();` model for calling the async method in a synchronous way
- Don't update the ExecuteQueryRetry methods 

See https://github.com/SharePoint/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/ALM/AppManager.cs for more details on the approach.

baywet:
Hey @jansenbe Thanks for the follow up!
The only difficulty with the current implementation in ClientContextExtensions is that is doesn't provide a ExecuteQueryRetryAsyncMethod, forcing conditional compilation all the way up.
That's going to add a lot of complexity and overhead. I'm going to propose a solution based on your implementation and let me know what you think.

baywet:
there https://github.com/SharePoint/PnP-Sites-Core/pull/1653/commits/7d1ab10f102e4068e4f1a5f4cfdd0ee634715895

baywet:
Alright I aligned the ExternalSharingExtensions, let me know if that matches what you expected. If yes, I'll update the Branding Extensions and we should be ready to merge this PR :)

jansenbe:
Hi Vincent, 

I manually merged the ExternalSharingExtensions.cs file and credited you for the work via the changelog. Looks really good, only minor change I did is instead of using **Base(method)** as the name of the private async method with the core logic I renamed these to be like **(method)Implementation** which is more aligned to how we named things across pnp sites core. Will close this PR for now, would recommend submitting future changes (e.g. async branding extensions) as one PR per extensions file. 

Thanks for the good contribution!

baywet:
Thanks @jansenbe ! I will open other pull requests. hopefully now that things are sorted out it should help streamline other contributions.

PR #1667, opened 2018-03-14
add SynchronizationContextRemover to ExecuteQueryImplementation since…

### LINKED ISSUE #1617: Deadlocks everywhere in PnP Core!!

OliverZeiser:
… this one is using ExecuteQueryAsync now

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no - yes?
| Related issues?  | fixes #1617 
Fixed the new async ExecuteQueryImplementation to work with the new SynchronizationContextRemover

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
@@ -100,6 +100,7 @@ namespace Microsoft.SharePoint.Client
         private static void ExecuteQueryImplementation(ClientRuntimeContext clientContext, int retryCount = 10, int delay = 500, string userAgent = null)
 #endif
         {
+            await new SynchronizationContextRemover();
             var clientTag = string.Empty;
             if (clientContext is PnPClientContext)
             {


jansenbe:
Hi @OliverZeiser ,

The SynchronizationContextRemover is called from the main public method that then calls the ExecuteQueryImplementation private method. So there's not need to call it again in my view. Do you things not working without it?

jansenbe:
FYI: I did try to use the provisioning engine from both a forms and a WPF app and got deadlocks...with the current dev version these resolved in my test scenario, but would be good if you could test your past deadlock scenarios as well

OliverZeiser:
It is definitly not working in the current DEV Branch. I am using a MVC Application. I have created a new Pull Request that shows the change that is needed. With this change it is working for me. Maybe you can test that again...
The problem is only with the ExecuteQueryAsync now that got added. It looks like you forgot to add the new SynchronizationContextRemover  there. But this leads to big issues, since PnP calls the ExecuteQueryAsync now in the ExecuteQueryRetry via the ExecuteQueryImplementation which is used everywhere.
You have only added the SynchronizationContextRemover  to the ExecuteQueryRetryAsync.

jansenbe:
Got it...did push in a change for this just now, meaning I'll close this PR. Thanks for the testing and let me know if you still see deadlocks in your testing

biste5:
you didn't take my suggestion @jansenbe 
https://github.com/SharePoint/PnP-Sites-Core/commit/5dc3e51ebb66c4c55f6f059a2dae0de03a9891ed#r27972838

:(

jansenbe:
@biste5 : happy to review your suggestion if you can indicate the current approach is still resulting in deadlocks or whether you think something is fundamentally wrong

biste5:
@jansenbe 

There's extra overhead (tasks, and reference types in memory) generated at
[ClientContextExtensions.cs#L73-L78](https://github.com/SharePoint/PnP-Sites-Core/blob/dev/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs#L73-L78)

should just be
```c#
public static Task ExecuteQueryRetryAsync(this ClientRuntimeContext clientContext, int retryCount = 10, int delay = 500, string userAgent = null)
{
    return ExecuteQueryImplementation(clientContext, retryCount, delay, userAgent);
}
```

jansenbe:
That's indeed cleaner and better. Did some testing with your code and did not see issues and will check-in this change. Would be good if @OliverZeiser could validate this scenario still works after this change

PR #1708, opened 2018-04-05
Fixed to set modern page layout property during page pre-creation

TheJeffer:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes regression introduced in fix 8717070b

Fixes regression where setting Layout attribute of ClientSidePage to Home isn't working. All modern pages are created with Article layout. Happens because as of now, layout of modern pages can't be switched from Article to Home or vice versa once the page is created. This means layout must be set during the page pre-creation process.


jansenbe:
Nice catch Jeff. Have merged this manually but mentioned you in changelog so you'll be mentioned in the release notes of May. Thx!

TheJeffer:
Sounds good. Thanks!

PR #1711, opened 2018-04-06
Handling ServerException from ServerErrorCode instead of Message in S…

### LINKED ISSUE #1705: "The object specified does not belong to a list" SP Language conflict

#### Category
[ x] Bug
[ ] Enhancement

#### Environment
[x ] Office 365 / SharePoint Online
[x ] SharePoint 2016
[x ] SharePoint 2013

_If you are reporting a bug, please describe the expected behavior. If you are suggesting an enhancement please
describe thoroughly the enhancement, how it can be achieved, and expected benefit._

When adding properties to the homepage file while applying provisioning template via powershell command "Apply-PnPProvisioningTemplate" we expect the site to be created with that template independently of our SP Language.

#### Observed Behavior
_If you are reporting a bug, please describe the behavior you expected to occur when performing the action. If you are making a suggestion, you can delete this section._

When the execution arrives to SetFileProperties method in ObjectFiles.cs it throws an exception with message "The object specified does not belong to a list." which sould be skipped according to the code because the actual exception message is the same but in spanish, which is the language of our SP site. There is a picture with the execution at this point.

image

#### Steps to Reproduce
_If you are reporting a bug please describe the steps to reproduce the bug in sufficient detail to allow testing. Only way to fix things properly, is to have sufficient details to reproduce it. If you are making a suggestion, you can delete this section._

Just try the command Apply-PnPProvisioningTemplate with a SP site in spanish (I guess there is the same problem in any language except english)

Ralmenar:
Bug fix
Related issues?  | fixes #1705 

 Handling ServerException from ServerErrorCode instead of Message in SetFileProperties method of File ObjectFiles.cs.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectFiles.cs
@@ -263,7 +263,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 catch (ServerException ex)
                 {
                     // If this throws ServerException (does not belong to list), then shouldn't be trying to set properties)
-                    if (ex.Message != "The object specified does not belong to a list.")
+                    if (ex.ServerErrorCode != -2146232832)
                     {
                         throw;
                     }


jansenbe:
Good catch. Will merge

PR #1755, opened 2018-05-11
Fix to support creating list from a list definition (stp)

wobba:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

Fix so that list instances created from list definitions (stp) files get's created correct. When a FeatureId is present we need to check the template and set `ListTemplate` instead of `TemplateType`.

Sample template
```xml
<?xml version="1.0"?>
<pnp:Provisioning 
    xmlns:pnp="http://schemas.dev.office.com/PnP/2018/05/ProvisioningSchema">
    <pnp:Preferences Generator="OfficeDevPnP.Core, Version=2.27.1806.0, Culture=neutral, PublicKeyToken=null" />
    <pnp:Templates ID="CONTAINER-TEMPLATE-B417026045BA442685C3AB4351F0BF21">
        <pnp:ProvisioningTemplate ID="TEMPLATE-B417026045BA442685C3AB4351F0BF21" Version="1" BaseSiteTemplate="STS#0" Scope="RootSite">
            <pnp:Lists>
                <pnp:ListInstance Title="PnP" TemplateType="100" Url="Lists/PnP" TemplateFeatureID="00bfea71-de22-43b2-a848-c05709900100"></pnp:ListInstance>
            </pnp:Lists>
        </pnp:ProvisioningTemplate>
    </pnp:Templates>
</pnp:Provisioning>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Changed
 - ClientComponentId and ClientComponentProperties are now updated when applying a template to a site where the customaction already exists [SchauDK]
 - Fixes issue with requiring tenant admin access while not provisioning tenant scoped artifacts
+- Fixed issue where a list would not be created based on a list template (TemplateFeatureId)
 
 ### Deprecated
 
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -1534,17 +1534,33 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
             }
             else
             {
-                var listCreate = new ListCreationInformation();
-                listCreate.Description = parser.ParseString(templateList.Description);
-                listCreate.TemplateType = templateList.TemplateType;
-                listCreate.Title = parser.ParseString(templateList.Title);
+                ListCreationInformation listCreate =
+                    new ListCreationInformation
+                    {
+                        Title = parser.ParseString(templateList.Title),
+                        Description = parser.ParseString(templateList.Description),
+                        Url = parser.ParseString(templateList.Url),
+                        // the line of code below doesn't add the list to QuickLaunch
+                        // the OnQuickLaunch property is re-set on the Created List object
+                        QuickLaunchOption = templateList.OnQuickLaunch ? QuickLaunchOptions.On : QuickLaunchOptions.Off
+                    };
 
-                // the line of code below doesn't add the list to QuickLaunch
-                // the OnQuickLaunch property is re-set on the Created List object
-                listCreate.QuickLaunchOption = templateList.OnQuickLaunch ? QuickLaunchOptions.On : QuickLaunchOptions.Off;
-
-                listCreate.Url = parser.ParseString(templateList.Url);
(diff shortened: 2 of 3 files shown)


sivamsolai:
Hi @wobba 

https://github.com/SharePoint/PnP-PowerShell/issues/1613

The code changes opened the other issue. During site Provisioning if more than one "TemplateFeatureID" is matched with List Template Gallery, got error message "Sequence contains more than one matching element". 
 
Suppose TemplateFeatureID is matched with one, The code creating list from list definition (stp) instead of PnP XML


image

wobba:
Ok.. so seems you can have the same feature id for multiple templates. I'll look at it.

wobba:
@sivamsolai Do you have a sample xml file which breaks? And I only need the `pnp:list` list element(s).

sivamsolai:
Hi @wobba 
Thanks for your reply 

[CustomListSchema.txt](https://github.com/SharePoint/PnP-Sites-Core/files/2349132/CustomListSchema.txt)

Please follow the steps to reproduce the issue 
1) Save the custom list as a list template with content      
2) Apply the site provisioning to sub site 

if the root site have more than one list template ,will get the ""Sequence contains more than one matching element"." error 

suppose the  root site have a one list template, the new list will be create  using list template  instead of pnp:list list element(s).

wobba:
Could you pull out a list of the custom templates on your site which you are running. Question is, when should it use an stp and when not. Need to find the right conditions to fix the code.

$site = Get-PnPSite
$web = Get-PnPWeb
$listDefs = $site.GetCustomListTemplates($web)
$web.Context.Load($listDefs)
Invoke-PnPQuery
$listDefs

sivamsolai:
Hi @wobba 

[ListCollection.txt](https://github.com/SharePoint/PnP-Sites-Core/files/2353273/ListCollection.txt)

attached the list collection details from site.

wobba:
@sivamsolai Looking at your template and looking at the stp list, what should the logic be if I wanted to target one of the stp's? Both template type (100) and feature I'd are the same. Would you expect one of them to be used as a template for your customer list, or none?

Maybe we need an stp attribute in the pnp schema..and if more than one stp matches, pick none. @PaoloPia or @erwinvanhunen do you have any suggestions?

sivamsolai:
@wobba 
 my suggestion also same ,  we need add an stp attribute in the pnp schema. When a stp attribute  is present, the code needs to pick the template from List Template Gallery, otherwise needs use "TemplateFeatureID" from  pnp:list list element(s).

wobba:
@PaoloPia @erwinvanhunen I have added an intermediate fix until we can decide how to handle stp's - as they seem have the same feature id and template type id.

sivamsolai:
@wobba 
could you please share the update for  stp's in pnp:list list element(s).

> @PaoloPia @erwinvanhunen I have added an intermediate fix until we can decide how to handle stp's - as they seem have the same feature id and template type id.

@wobba 
could you please share update for the  above issue.

wobba:
@sivamsolai See https://github.com/SharePoint/PnP-Sites-Core/commit/cdc8103d79ec473bec08487a7a5616c737294500 for my intermediate fix which was included in the September release.

sivamsolai:
@wobba 
The list need to be create from pnp:list list element(s).  suppose if root site has list definitions (stp) files the code is template, the code is creating list from. stp file. if .stp file saved with "Include content", new list created with content

wobba:
@sivamsolai Currently if you have one stp with a unique id, that is used. I suggest moving off stp’s and using the pnp:datarow option instead, at least until someone writes proper stp code with a change in the schema to support this.

Issue #1656, opened 2018-03-08
Export-PnPTermGroupToXml does not export custom sort order if site collection termgroup


metagic:
#### Category
[x] Bug

#### Environment
[x] Office 365 / SharePoint Online

#### Expected or Desired Behavior
The exported template should include the custom sort order.

#### Observed Behavior
For site collection termgroups, term ids are now nullified (this seems to be a new behaviour in the March 2018 release of the cmdlets). Therefore, `var term = termsToReturn.FirstOrDefault(t => t.Id == Guid.Parse(id));` in ObjectTermGroups.cs will _always_ return null and no custom sort order is added to the exported term definition.

PR #1761, opened 2018-05-14
Fix issue #1656

TeodoraI:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

Fix issue #1656
There was a condition to set the term id only if it not site collection term or it is reused:
`if (!isSiteCollectionTermGroup || term.IsReused)
                {
                    modelTerm.Id = term.Id;
                }`
However, if the id of the term is not set, then later the custom sort order won't be returned for site collection terms because it is requested by id:
`                    var term = termsToReturn.FirstOrDefault(t => t.Id == Guid.Parse(id));
                    if (term != null)
                    {
                        term.CustomSortOrder = count;
                        count++;
                    }`


PaoloPia:
Hi @TeodoraI,
Thanks for your feedback. Actually, we skip exporting the term ID for a term in a Site Collection Term Group, to avoid having duplicate IDs across different Site Collections, because it would lead to an exception while importing the terms. Have you tried this change exporting and importing the same Site Collection Term Group across different Site Collections within the same tenant? Thanks.

TeodoraI:
Hi Paolo,
Sorry for the late reply. I tested it and you are right. I returned the check for !isSiteCollectionTermGroup and I changed the logic of adding the custom sort order so it will be exported even if the term id is not filled in. Thanks!

PaoloPia:
Hi @TeodoraI,
Sorry for the long wait. The code suggested looks now ok to me, and I just merged it.

PR #1820, opened 2018-06-14
bugfix: removing unecessary task.run for the synchronous methods

baywet:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

- Removed unecessary Task.Run usage in synchronous methods to make code maintainability better and to avoid spawning unecessary tasks (reduced memory usage)

Hi @jansenbe 
Reviewing my recent submissions around the async work I noticed some had a task.run wrapper and some had not.  
I think this was a relic of the early discussions we had for async support and context sync.  
I ran a little POC console application on my side Task.Run is supposed to defer the work on a thread pool but because we're using GetAwaiter().GetResult() right away we're blocking the current thread anyway.  
The idea here is to remove some proxy tasks and thread management overhead for synchronous work. That's also going to make the code maintenance easier for us.  
https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=netframework-4.7.1#System_Threading_Tasks_Task_Run_System_Func_System_Threading_Tasks_Task__

### DIFF
--- a/Core/OfficeDevPnP.Core/ALM/AppManager.cs
+++ b/Core/OfficeDevPnP.Core/ALM/AppManager.cs
@@ -55,7 +55,7 @@ namespace OfficeDevPnP.Core.ALM
         /// <returns></returns>
         public AppMetadata Add(byte[] file, string filename, bool overwrite = false, AppCatalogScope scope = AppCatalogScope.Tenant, int timeoutSeconds = 200)
         {
-            return Task.Run(() => AddAsync(file, filename, overwrite, scope, timeoutSeconds)).GetAwaiter().GetResult();
+            return AddAsync(file, filename, overwrite, scope, timeoutSeconds).GetAwaiter().GetResult();
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


jansenbe:
Thanks @baywet. I'll have a look into this...I think added the task run to avoid hangs when being used from within a "old school" Windows Forms app...however I'll verify again. So you did verify this from what type of applications? ASP.Net MVC? WPF?

baywet:
Console classic and console dot net core so far

jansenbe:
Hi Vincent, I did some testing and without Task.Run(() the methods dead lock in a windows forms app...hence I've added Task.Run(() =>  where it was not there yet and will close this pr.

baywet:
Hi @jansenbe 
Ok thanks the taking the time to test. Interesting to see that the Task Scheduler behaves this way in a windows forms app.
I need to submit another PR then to fix the ones I forgot.
For the testing apps you're using, do you have those somewhere on a repos?
Thanks!

jansenbe:
I fixed the file/folder ones already...fix program is not (yet) shared...it's a little GUI that internal teams use to extract/apply pnp provisioning templates, will see if I can somehow share that one.

baywet:
ok thanks! no work required for me on this front then, I'll send you a PR for new async methods instead.

PR #1821, opened 2018-06-14
Fix NullReferenceException when parsing client side page header html

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | NA

When parsing the Client Side Page header Html, some properties might not exist within the html (eg. siteId, webId), which will result in a NullReferenceException.
I've added null check for the properties.


jansenbe:
Thanks Michael!

PR #1822, opened 2018-06-14
Update FieldUtilities.cs

cebud:
Changed multi lookup field privisioning to also handle list url in `List`

| Q               | A
| --------------- | ---
| New feature?    | yes

Changed multi lookup field privisioning to also handle list url in `List`


jansenbe:
Thanks @cebud , good change.

PR #1846, opened 2018-07-06
Stability improvements for updates to RoleDefinition update

sebastianmattar:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

I encountered a strange behaviour with SharePoint 2013 that updates to BasePermissions did not work. The solution is to create a new BasePermission object, assign it to the RoleDefinition and update.

Additionally I made a small optimization: the IsDirty check now considers existing permissions as well as the permissions in the template.


sebastianmattar:
Any news on when this fix can be merged?

PR #1851, opened 2018-07-10
Fix - UpdateUnifiedGroup method - invalid response header issue, …

### LINKED ISSUE #1807: Invalid property 'responseHeaders' when call UnifiedGroupsUtility.UpdateUnifiedGroup()
#### Category
[x ] Bug

#### Environment
[x ] Office 365 / SharePoint Online

[UnifiedGroupsUtility.UpdateUnifiedGroup()](https://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Graph/UnifiedGroupsUtility.cs) updates group properties (display name, description, visibility, owners, members)

#### Observed Behavior
The following error is returned when try to update e.g. description:
`
HTTP/1.1 400 Bad Request

ef
{
  "error": {
    "code": "Request_BadRequest",
    "message": "Invalid property 'responseHeaders'.",
    "innerError": {
      "request-id": "...",
      "date": "..."
    }
  }
}
`

#### Steps to Reproduce
Call this method for some group:
`UnifiedGroupsUtility.UpdateUnifiedGroup(group.GroupId, accessToken, description: "test");`

#### Submission Guidelines
This error started to be shown recently. Before it worked Ok. The problem is caused by .Net graph client library which is used inside UnifiedGroupsUtility.UpdateUnifiedGroup(): because of some reason it sends responseHeaders property to HTTP PATCH method among with other group's properties. There is workaround for this problem: construct Group object with minimal required properties which should be updated:
`
var graphClient = CreateGraphClient(accessToken);
var existingGroup = await graphClient.Groups[groupId].Request().GetAsync();

var groupToUpdateMinimal = new Group();
groupToUpdateMinimal.Id = groupId;

bool updateGroup = false;
if (!string.IsNullOrEmpty(description) && existingGroup.Description != description)
{
	groupToUpdateMinimal.Description = description;
	updateGroup = true;
}

// Check if visibility has changed for the Group
bool existingIsPrivate = existingGroup.Visibility == "Private";
if (isPrivate != null && existingIsPrivate != isPrivate.Value)
{
	groupToUpdateMinimal.Visibility = isPrivate.Value ? "Private" : "Public";
	updateGroup = true;
}

if (updateGroup)
{
	await graphClient.Groups[groupId].Request().UpdateAsync(groupToUpdateMinimal);
}
`
In this case only these updated properties will be send to Graph API with HTTP PATCH request:
`
{
  "description": "test",
  "visibility": "Public",
  "id": "..."
}
`
and group will be update successfully.

gautamdsheth:
…toggle group visibility bug

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #1807 

1) Fixes #1807  - invalid response headers while making an update to a unified group
2) If the Unified's group visibility was set to Private and we try to update the group by just changing the description, the group's visibility was getting set to Public because the parameter `isPrivate` was set to false. Fixed it by making it nullable

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Graph/UnifiedGroupsUtility.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Graph/UnifiedGroupsUtility.cs
@@ -389,7 +389,7 @@ namespace OfficeDevPnP.Core.Framework.Graph
             var fullListOfOwners = await graphClient.Groups[targetGroup.Id].Owners.Request().Select("userPrincipalName, Id").GetAsync();
             var pageExists = true;
 
-            while(pageExists)
+            while (pageExists)
             {
                 foreach (var owner in fullListOfOwners)
                 {
@@ -444,7 +444,7 @@ namespace OfficeDevPnP.Core.Framework.Graph
         public static bool UpdateUnifiedGroup(string groupId,
             string accessToken, int retryCount = 10, int delay = 500,
             string displayName = null, string description = null, string[] owners = null, string[] members = null,
-            Stream groupLogo = null, bool isPrivate = false)
+            Stream groupLogo = null, bool? isPrivate = null)
         {
             bool result;
             try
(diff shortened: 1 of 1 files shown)


PaoloPia:
Hi @gautamdsheth,
Thanks for your submission. This PR has been merged.

Issue #1461, opened 2017-11-19
Default Column values not applied in Libraries provisioned via Apply-PnPProvisioningTemplate


erwinvanhunen:
_From @sierra710 on October 20, 2017 20:2_

###Notice: many issues / bugs reported are actually related to the PnP Core Library which is used behind the scenes. Consider carefully where to report an issue:###

1. **Are you using ```Apply-SPOProvisioningTemplate``` or ```Get-SPOProvisioningTemplate```**? The issue is most likely related to the Provisioning Engine. The Provisioning engine is _not_ located in the PowerShell repo. Please report the issue here: https://github.com/officedev/PnP-Sites-Core/issues.
2. **Is the issue related to the cmdlet itself, its parameters, the syntax, or do you suspect it is the code of the cmdlet that is causing the issue?** Then please continue reporting the issue in this repo.
3. **If you think that the functionality might be related to the underlying libraries that the cmdlet is calling** (We realize that that might be difficult to determine), please first double check the code of the cmdlet, which can be found here: https://github.com/OfficeDev/PnP-PowerShell/tree/master/Commands. If related to the cmdlet, continue reporting the issue here, otherwise report the issue at https://github.com/officedev/PnP-Sites-Core/issues

### Reporting an Issue or Missing Feature
After provisioning multiple subsites with Apply-PnPProvisioningTemplate, when testing a library that was created via the template.xml file, it was observed that default column values (List columns not Site columns) are not applied when files are uploaded via OD4B or File Explorer (Copy/Paste) or Drag/Drop to the library, however the default column values are applied when uploading a file from the Ribbon.

### Expected behavior 
that default column values will be applied regardless of how the files are uploaded

### Actual behavior 
Default column values are not applied

### Which version of the PnP-PowerShell Cmdlets are you using?
- [ ] PnP PowerShell for SharePoint 2013
- [ ] PnP PowerShell for SharePoint 2016
- [x] PnP PowerShell for SharePoint Online

### What is the version of the Cmdlet module you are running?
SharePointPnPPowerShellOnline 2.19.1710.1

### How did you install the PnP-PowerShell Cmdlets? 
- [ ] MSI Installed downloaded from GitHub
- [x] Installed through the PowerShell Gallery with Install-Module
- [ ] Other means

_Copied from original issue: SharePoint/PnP-PowerShell#1168_

erwinvanhunen:
_From @sierra710 on October 20, 2017 20:4_

[...]

PR #1862, opened 2018-07-26
Default column values in subsites

cnesmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | Might fix issue #1461 - but this has not been tested.

## A bug fix for using `SetDefaultColumnValues` in lists in subsites.

When the method `Microsoft.SharePoint.Client.ListExtensions.SetDefaultColumnValues` is used to set default column values in a list in a subsite (not rootWeb), an error is thrown, "File not found."

The root cause for this is found in the private method `GetFormsFolderFromList`, which uses `clientContext.Web.GetFolderByServerRelativeUrl`. In other words, the root web is always queried. But when the list in question resides in a subsite, `GetFolderByServerRelativeUrl` returns "File not found" since the root web does not contain a library with the URL _/my-subsite/Pages/Forms_.

The code has been changed to `list.ParentWeb.GetFolderByServerRelativeUrl`, and now works with subsites, too.


jansenbe:
Thanks for the fix Christian.

Issue #1812, opened 2018-06-11
List and site Lookup fields no longer working


waaromikniet:
#### Category
[x] Bug
[ ] Enhancement

#### Environment
[x] Office 365 / SharePoint Online
[x] SharePoint 2016
[x] SharePoint 2013

#### Expected or Desired Behavior
In my template I have a list lookup field like below. I expect this to be provisioned. This works with the January 2018 release.

`            <Field ID="{fa564e0f-0c70-4ab9-b863-0177e6ddd247}" Name="Title" Required="true" Indexed="TRUE" EnforceUniqueValues="TRUE" Type="Text" StaticName="Title"/>
            <Field Type="Lookup" List="Lists/DocumentExchangeLibrary" Name="DocumentProcessStatus" StaticName="DocumentProcessStatus" DisplayName="Document Process Status" Group="DCP" ShowField="CalStatus" Indexed="TRUE" Hidden="TRUE" Viewable="FALSE" ID="{7f5ad3ab-94dc-4999-8f03-7d5a701c8c49}"/>`

#### Observed Behavior
The above field XML is no longer working. It seems that the attribute List="Lists/DocumentExchangeLibrary"  is no longer working with the May release.

#### Steps to Reproduce

This field XML does not work

`            <Field ID="{fa564e0f-0c70-4ab9-b863-0177e6ddd247}" Name="Title" Required="true" Indexed="TRUE" EnforceUniqueValues="TRUE" Type="Text" StaticName="Title"/>
            <Field Type="Lookup" List="Lists/DocumentExchangeLibrary" Name="DocumentProcessStatus" StaticName="DocumentProcessStatus" DisplayName="Document Process Status" Group="DCP" ShowField="CalStatus" Indexed="TRUE" Hidden="TRUE" Viewable="FALSE" ID="{7f5ad3ab-94dc-4999-8f03-7d5a701c8c49}"/>`

When changing List="Lists/DocumentExchangeLibrary" to List="{listid:Document Exchange Library}" WebId="{siteid}"

my template does work. Is  List="Lists/DocumentExchangeLibrary"  no longer supported?

waaromikniet:
#1786 # #1791

ezybzy:
For me, it fails with the following field schema.
[...]

Issue #1786, opened 2018-05-23
Provisioning: Lookup field not provisioned correctly in May release


mkacena:
#### Category
[ x ] Bug
[ ] Enhancement

#### Environment
[ x ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

#### Expected or Desired Behavior
Provision a Lookup field correctly.

#### Observed Behavior
After upgrading the NuGet to version 2.26.1805.1, the provisioning engine does not correctly provision fields of type Lookup. When the provisioning is complete, the site column's reference to the list is empty and trying to add the column to any list ends with an error.

Running the provisioning with version 2.25.1804.1 provisions the field correctly.

#### Steps to Reproduce
Create a provisioning template with field definition such as this one and provision the template to a web:

```
<Field
          ID="{3E1AA947-9A9A-4F8D-9E31-8BF08FC31186}"
          Name="ProEntityPicker"
          DisplayName="Entity"
          Type="Lookup"
          List="Lists/Entities"
          Required="FALSE"
          EnforceUniqueValues="FALSE"
          ShowField="Title"
          UnlimitedLengthInDocumentLibrary="FALSE"
          Group="{localization:Group_Dms}"
          Overwrite="TRUE"
          SourceID="{parameter:SourceId}">
</Field>
```

jimmywim:
We've also noticed this issue. Reverting to 2.25 works.

jansenbe:
We'll have a look at this. Adding @stevebeauge  for awareness.

stevebeauge:
I repro-ed the issue when the `List` attribute is not a GUID but the web relative Url of the target list. I'll take a look.

jansenbe:
@mkacena  / @jimmywim : Please test with the June release, will be out really soon

mkacena:
Tested the June release 2.27.1806 and it still does not fully behave as expected:

**Scenarios**

1. Clean deployment to a single web

Result: OK

2. Corrective deployment to a single web

Result: Not working
Setup: I had a Lookup field provisioned with bad release 2.26.1805.1. The field was corrupted. After upgrading to release 2.27.1806 and repeating the provisioning, the field did not get fixed. This worked with previous release 2.25.1804.1.

3. Deployment to a site collection

Result: Not working
[...]

PR #1872, opened 2018-08-03
Bugfix/lookupfield

antim-mironov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | partially #1812, mentioned in #1786

The PR fixes an issue with lookup fields in a list instance, when a template is applied to update a lookup field. 
The PR implements also new tests to validate this case.
The test "ObjectListInstanceTests->CanUpdateLookupFieldLocallyInListInstance" would fail if the bug fix is not applied.


jansenbe:
Thx Antim, good change which will be merged just in time for the next release :-)

PR #1895, opened 2018-09-06
Fix: LoginNames should be compared case insensitive

### LINKED ISSUE #1881: LoginNames should be compared case insensitive
#### Category
[x] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[x] SharePoint 2013

If SharePoint on-premises, what's exact CU version: CU January 2018

Sucessfully assigning a `<RoleAssignment>` during provisioning of a List via the `<ListInstance><Security /></ListInstance>` element regardless of the of the casing of the name.

#### Observed Behavior
If the group name in the provisioning template is not exactly cased like the LoginName property of the group in the SharePoint SiteGroups collection. The provisioning engine will try to create a new User via EnsureUser which will fail with an exception because the account obviously cannot be found.

#### Steps to Reproduce

1. Create a group inside a SharePoint Site
2. Create a provisioning template which creates a `<ListInstance />` and assignes special `<Security>` by referencing the group created in (1.) Use another casing in the provisioning template than in the LoginName of the group.
3. Apply the template
4. An exception will be thrown stating that the account cannot be found.

####  Proposed resolution

Find attached a patch file which changes the comparison of LoginNames from the equality operator '==' to .Equals using StringComparison.OrdinalIgnoreCase.

tmeckel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #1881 

As described in Issue #1881 the comparison of LoginNames should be performed case-insensitive. 
The PR contains all changes to the master branch to change the comparison logic.

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/TokenParserTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/TokenParserTests.cs
@@ -16,13 +16,13 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
         {
             using (var ctx = TestCommon.CreateClientContext())
             {
-                ctx.Load(ctx.Web, 
-                    w => w.Id, 
-                    w => w.ServerRelativeUrl, 
+                ctx.Load(ctx.Web,
+                    w => w.Id,
+                    w => w.ServerRelativeUrl,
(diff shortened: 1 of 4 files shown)


jansenbe:
Thanks @tmeckel , this is good change that we'll merge in.

PR #1958, opened 2018-11-01
Decode filename before upload

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

If a persisted file (eg. Site Logo) contains a space within the filename, it cannot be uploaded when the template is applied as the UploadFile method will throw an ArgumentException.

This small fix will UrlDecode the filename before uploading.


jansenbe:
Hi @SchauDK, the proposed behavior is already present as a fallback mechanism...not going to change that now to avoid possible backward compatibility issues.

SchauDK:
@jansenbe The folder extension UploadFile is throwing an ArgumentException.
The fallback is to catch an ServerException, so the fallback doesn't work in this situation.

In my opinion it is also better to handle it (decode) instead of throwing an exception, catch and then decode and retry.

PR #1968, opened 2018-11-09
Fixed token parsing and removed ~ tokens tests as they are not supported anymore (#1953)

phawrylak:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #1953

#### What's in this Pull Request?

In Workflow XAML definition `{}` is used as escape sequence. In case described in #1953, we have `{}{0}Id` in one of attributes. After token parsing it is changed to `{0}Id` which is wrong. It is effect of a bug in `TokenParser.ParseString()` method. This PR fixes it and removes ~ tokens tests as they are not supported anymore.

PR #1972, opened 2018-11-09
BuildTokenCache bug

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

BuildTokenCache is storing values in the TokenDictionary with an unescaped key, but checking using a key which is not unescaped.

Furthermore ObjectListInstance.ProcessFieldRefs has become extremely slow as the LINQ is calling ParseString. ParseString is removed as I cannot see that it should be necessary to just get the ID attribute.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -389,7 +389,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 var fieldsRefsToProcess = listInfo.TemplateList.FieldRefs.Select(fr => new
                 {
                     FieldRef = fr,
-                    TemplateField = template.SiteFields.FirstOrDefault(tf => (Guid)XElement.Parse(parser.ParseString(tf.SchemaXml)).Attribute("ID") == fr.Id)
+                    TemplateField = template.SiteFields.FirstOrDefault(tf => (Guid)XElement.Parse(tf.SchemaXml).Attribute("ID") == fr.Id)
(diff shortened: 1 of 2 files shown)


wobba:
I'll include the tokenparser changes in a manual merge with #1968 and some other cleanup. See https://github.com/wobba/PnP-Sites-Core/tree/tokenimprov for now.

And we cannot remove `parseString` for the id, as some templates are using a token as the id.

The refactorings by @stevebeauge which solved an important issue did decrease perf a lot, especially for larger templates - which I managed to speed up again quite a bit with the refactoring of how the token parsing now works.

If you have suggestions to speed up Linq, that be local/global caching, pre-checking etc, I'm all for it.

Do you have benchmarks on how much that one `parseString` impacts a particular template?

wobba:
Manually merged with #1975

SchauDK:
Thanks @wobba 

I've done a benchmark for a template with 16 lists.
The average execution time per iteration is
Without parseString: 13ms
With parseString: 198215ms

So it is REALLY bad.
I'll come up with at suggestion to speed up the LINQ.

wobba:
@SchauDK I totally see why it happens, can you pull down the latest branch and see how much my latest commit helped? Added a guid check to avoid iterating all tokens for a match.

SchauDK:
@wobba It's for sure better.
Avg. execution time is now down to 13946ms.

What do you think about this change to completely remove the performance issue within ProcessFieldRefs?
https://github.com/SchauDK/PnP-Sites-Core/tree/ProcessFieldrefsImprovement

wobba:
@SchauDK ..if you create a PR with that branch I can have a look at getting it in.

SchauDK:
@wobba Created PR. Sorry, haven't had time to add tests.

PR #1974, opened 2018-11-10
Extract MUI for List Views and Navigation Nodes / fix MUI List View Provisioning

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes 
| New feature?    |  yes
| New sample?      | no 

Bug:
ObjectListInstances =>CreateView  did not handle MUI Tokens since SchemaXml was parsed.ParseString and therefore ContainsResourceToken would never be true and LocalizeView will not be calledt.

New Feature:
Since already looking at the code i added Extract MUI View Titles and MUI Navigation Nodes

Tested with Englisch/German/Italian => Once Web Language EN and once Web Base Language DE


czullu:
Hi, what kind of investigation is needed? @PaoloPia

PR #2018, opened 2018-12-20
Log level logging fixes

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fixed logic to decide if logging should happen based on configured log level. Log levels with equal or higher severity than the configured log level will now always be logged.
Previously the behaviour was that when logging an error it would only log the event if log level was set to error or debug. Errors should always be logged. Similar issues existed when logging information (would log if log level was set to higher severity than information).


jansenbe:
Thx @jensotto, this is a good change!

PR #2020, opened 2018-12-20
Provision and extract associated groups

jensotto:
…member/visitor groups

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

Added support for provisioning and extracting associated owner, member, and visitor group.

I have changed the creation of groups in ObjectSiteSecurity to not require a group owner. This is not enforced by SharePoint, and will result in the current user to be the group owner. This needs to be changed in the schema too to be fully supported.

Changed group creation to support clearing description if it's an empty string in the schema.

I believe ClientObjectExtensions.ServerObjectIsNull is incorrectly implemented. I added a null check as the method could throw null reference exception. It also returns true when ClientObject.ServerObjectIsNull is null. This is undefined and should not return true as it can not be determined if the server object in fact is null. In order to not change the behaviour of this public method I was forced to create a new one in PnPUtility.

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
@@ -6,6 +6,7 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using OfficeDevPnP.Core.Entities;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
 using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers;
+using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.TokenDefinitions;
 using OfficeDevPnP.Core.Utilities;
 using User = OfficeDevPnP.Core.Framework.Provisioning.Model.User;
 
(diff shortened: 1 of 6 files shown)


phawrylak:
I hope it will be merged for the next version as we really need to be able to set associated groups in template.

jansenbe:
Thx for this PR @jensotto, just merged it.

Regarding the ServerObjectIsNull: in your code you were checking this property before it was populated at the server. I've added an executequeryretry() to trigger the server to do the work (and populate the property). After that the default ServerObjectIsNull method worked fine in your scenario. I therefore have dropped your version of ServerObjectIsNull.

jensotto:
@jansenbe It likely will work, but it proves my point that the PnP version of ServerObjectIsNull is not working well. The change leads to an unnecessary round trip to the server. Every user of the ServerObjectIsNull also needs to ensure that a roundtrip to the server has been made before using that method. This means that if part of the code is building up a client side representation of the server object, a round trip needs to be made before ServerObjectIsNull can be used. In my contribution I create new objects depending on certain preconditions, and it would save a server roundtrip if ServerObjectIsNull returned false as expected when ClientObject.ServerObjectIsNull is null. It might not be a big performance hit for this single instance, but PnP Provisioning engine is becoming slower and slower and small adjustments can help preventing server roundtrip growth.
You can also see how Microsoft has implemented this for internal use in the asembly Microsoft.SharePoint.Taxonomy in method Microsoft.SharePoint.Taxonomy.Hybrid.RemoteService.IsClientObjectNull(ClientObject)
```
private static bool IsClientObjectNull(ClientObject clientObject) 
{
     if (clientObject == null)
     {
         return true;
     }
     if (clientObject.ServerObjectIsNull == null)
     {
         return false;
     }
     return clientObject.ServerObjectIsNull.Value;
 }
```
I reallise now that maybe the wording of the Method is the key here. Maybe renaming my contributed ServerObjectIsNull to ClientObjectIsNull would make it clearer? This is just a thought, but I still believe my version of the ServerObjectIsNull should be included for use inside PnP.

Also, I see that you removed my change for enabling the possibility to clear the group description. I believe this should be possible to do. Why would you not have that possibility in the engine?

jansenbe:
@jensotto : you're right about the group description change, I've reverted back to your original code for that.

About the ServerObjectIsNull: having it named as ClientObjectIsNull makes more sense as it makes no sense to verify a nullable value of a property that can only be set by the server before the call to the server has been made. Feel free to add a new extension method in the ClientObjectExtensions class.

About perf: I agree perf is far from optimal and work is needed to optimize...

Issue #2014, opened 2018-12-17
New throttling guidance is not correctly implemented in OfficeDevPnP


SchauDK:
#### Category
[X] Bug
[ ] Enhancement

#### Environment
[X ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

#### Expected or Desired Behavior
According to the new throttling guidance, we should read the "Retry-After" response header and use that for backing off before retrying a request.

In the OfficeDevPnP ExecuteQueryImplementation the Retry-After value of 120 is assigned the retryAfterInterval. This means that the thread will sleep 120ms instead of 120 seconds.
For each retry (default 10) the backoff interval will not increase.

I think you should revert back to use the "old" backoff structrure starting at 500ms and multiplying by 2 for each iteration. Waiting 2 minutes between each retry doesn't make sense, if this was the intention of the guidance change (besides using RetryQuery).

jansenbe:
Hi @SchauDK ,

The recommendation is to use the wait time provided by the service... the service should increase the wait time if needed. Do you see failures after being throttled with the returned timeout for max 10 times?

SchauDK:
Hi @jansenbe 

The Retry-After value is 120 (**in seconds**).
The Thread.Sleep(120) waits 120 milliseconds. I don't believe this is the intention??

Secondly, the Retry-After value will never change. At least I have never seen a value different from the default 120. 

So, in my case it is retrying 10 times and only waiting 120ms between the retries.

jansenbe:
If the retry-after is never changing then this indeed is not good. I'll check that first.

jansenbe:
Did talk with our devs and the header indeed is in seconds, I made a mistake with that one and will adjust

jansenbe:
Adjusted. Let me know if you're ok with closing the issue?

SchauDK:
I've reviewed your change.
There's a lot of code out there depending on this OfficeDevPnP library, so I think it's dangerous to change the delay parameter from milliseconds to seconds. If someone is using this parameter, the delay could potentially be VERY long.
Also, even though the official guidance says to use the response header (which is 120 seconds), I still think that this delay between each retry is long. I'd rather still use incremental delays. 500ms, 1000ms, 2000ms, 4000ms, 8000ms etc.

You also need to change the Log.Warning to log the correct backoffInterval.

[...]

PR #2021, opened 2018-12-21
Update ClientContextExtensions.cs

harlenalvarez:
The Retry-After header contains the back off time in seconds, so we need to transform them into ms if we received it.
Fixed exception message spelling

| Q               | A
| --------------- | ---
| Bug fix?        | yes?
| New feature?    | no?
| New sample?      | no?
| Related issues?  | fixes not waiting correct back off time

Converts the Retry-After value to ms as long as is a valid integer.


SchauDK:
@harlenalvarez See #2014

jansenbe:
This was fixed a few days ago, we've reverted back to specifying milliseconds as input to the call and we're not using the retry-header anymore. Thanks for your PR, will close it due to the updated implementation we already have in dev. If you feel different then feel free to comment and re-open the pr.

PR #2071, opened 2019-02-14
ALM functions for SharePoint 2019 on premises

lafe:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | partially https://github.com/SharePoint/PnP-PowerShell/issues/1812

As we needed the ALM functions in SharePoint 2019, I built upon the partially prepared code to include it. Changes are:

- Added SP2019 binaries from NuGet to the assemblies folder and updated the project file to use them if SP2019 is the target. I am not sure what is included in the 16.0.10000 folder as these files are from 2016 but are only used for the SP2019 branch.
- Updated the readme file for the assemblies to include new folder and updated the layout to follow Markdown best practices
- Fixed error in `RESTUtilities.SetAuthenticationCookies` for on premises environments, as the credentials passed are not of type `SharePointOnlineCredentials` but `NetworkCredentials`. In my tests, I did not need additional authentication data in my tests, but - depending on the authentication used in the target SharePoint - this might not be complete.
- The `GetAvailable*` tests in the `ALMTests` class assumed that at least one app is already in the App Catalog. I merged this check with the `AddRemove*` tests to be sure that at least one app has been added to the App Catalog before the methods are tested.


jansenbe:
Hi @lafe ,

Thanks for this usefull change. I want to merge this but would like to ask if you could update this PR so that the existing binaries are kept as is. These binaries are for SP2019 and we'll update these whenever needed. Is there an API that was not available in the existing binaries?

lafe:
Closing this PR in favor of #2073 and #2074. #2074 ensures that the now unnecessary binary blob will not be part of the history and therefore saves the storage space on all of our computers.

PR #2074, opened 2019-02-15
ALM functions for SharePoint 2019 on premises

lafe:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | partially SharePoint/PnP-PowerShell#1812, successor of #2071 

As we needed the ALM functions in SharePoint 2019, I built upon the partially prepared code to include it. Changes are:
- Fixed error in RESTUtilities.SetAuthenticationCookies for on premises environments, as the credentials passed are not of type SharePointOnlineCredentials but NetworkCredentials. 
- Added additional authentication data in `AppManager` and `ClientContextExtensions` for on premises scenarios with NetworkCredentials. This was sufficient for my scenarios, however, I am not sure, if it is completley sufficient for all possible scenarios.
- The GetAvailable* tests in the ALMTests class assumed that at least one app is already in the App Catalog. I merged this check with the AddRemove* tests to be sure that at least one app has been added to the App Catalog before the methods are tested.

This is the successor of PR #2071 that also included updated SP2019 assemblies. As this was not deemed to be necessary, I created a new pull request that does not contain the assemblies to remove the binary blob from the public history (otherwise, it would be forever part of the Git repository even though it is never used).

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/ALM/ALMTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/ALM/ALMTests.cs
@@ -1,4 +1,4 @@
-﻿#if !ONPREMISES
+﻿#if !SP2013 && !SP2016
 using Microsoft.Online.SharePoint.TenantAdministration;
 using Microsoft.SharePoint.Client;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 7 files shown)


jansenbe:
Excellent work @lafe ! Works perfect, thanks for this great contribution.

PR #2075, opened 2019-02-16
Set Group classification property using MS Graph .NET SDK

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | fixes NA

Changed the internals of  `SetSiteClassification` method to consume the Microsoft Graph .NET SDK.
This is quite similar to the implementation we have in the `UnifiedGroupsUtility`. 
Currently, since classification property can only be set via BETA endpoints, have continued using that, but simply changed to do that that via the SDK instead of the raw requests. 
This will make it easier in the future to simply move to the GA endpoint.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/SiteExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/SiteExtensions.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.SharePoint.Client;
+﻿using Microsoft.Graph;
+using Microsoft.SharePoint.Client;
 using OfficeDevPnP.Core.Framework.Graph;
 using System;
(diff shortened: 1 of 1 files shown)


PaoloPia:
Hi @gautamdsheth,
Thanks for your contribution. Actually, right now we're trying to make the PnP Sites Core library free from external libraries like the Graph SDK for .NET, for the sake of avoiding dependencies that can eventually become issues while upgrading the main library itself. As a consequence, we have a long term plan to completely remove the Graph SDK dependency from PnP Sites Core. Thus, even if your proposed change is good and totally ok from a syntactical and functional perspective, I'm sorry to reject and close it. Thanks.

PR #2077, opened 2019-02-16
Feature: Added support for special characters while filtering groups and users

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| New sample?      | no
| Related issues?  | fixes https://github.com/SharePoint/PnP-PowerShell/issues/1327  & https://github.com/SharePoint/PnP-PowerShell/issues/1272

1) Added support for filtering group's with special characters in the display name and mailnickname
2) Added support for special characters in user's principal name while updating owners and members.
3) Modified groups filter from `startsWith` to `eq` to fix that PnP PS issue (1272) while provisioning groups
4) Also, added a `count > 0` check while adding owners and members during groups provisioning process


wobba:
Thank you and will process this.

PR #2128, opened 2019-03-23
Fix : Skip creation of associated groups if they already exist

### LINKED ISSUE #2127: Associated groups from template should not be created
#### Category
- [x] Bug ?
- [x] Enhancement

#### Environment
- [ ] SharePoint 2016
- [ ] SharePoint 2013

3.7.1903.0ö

When provisioning from a template I don't expect the associated groups of the template (Owner, Members, Visitors) to be provisioned. If anything I expect the members to be mapped to the (already existing, differently named) associated groups of the new site.

#### Observed Behavior
The associated groups from the template are being created thus you have every group two times, but with different site names. One is something like "Site _template_ owners" and one is "Site _newsite_ owners".

#### Steps to Reproduce

1. export site as template using -All handler
2. import from template into a new site using -All handler
3. look at newsite/_layouts/15/groups.aspx
4. see all Owner, Member and Visitor groups two times

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| New sample?      | no
| Related issues?  | fixes #2127 

This PR fixes the issue of additional groups getting created while provisioning the template.
While debugging this, also found that everytime we create a new group and associate them, we need to run executeQueryRetry as well. 
Besides that also did a delta check to see if the Associated group name has changed in the groups, if it has changed then we will set that as the default group.

Have tested it quite exhaustively and it works fine for all sites , @jansenbe  can you have a look at this and let me know your thoughts ?

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
@@ -41,7 +41,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 
                 var siteSecurity = template.Security;
 
-                if (web.EnsureProperty(w => w.HasUniqueRoleAssignments))
+                if (web.EnsureProperty(w => w.HasUniqueRoleAssignments) && !web.IsNoScriptSite())
                 {
                     string parsedAssociatedOwnerGroupName = parser.ParseString(template.Security.AssociatedOwnerGroup);
                     string parsedAssociatedMemberGroupName = parser.ParseString(template.Security.AssociatedMemberGroup);
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
@@ -54,13 +54,34 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                         if (string.IsNullOrEmpty(parsedAssociatedOwnerGroupName))
                         {
                             web.AssociatedOwnerGroup = null;
+                            webNeedsUpdate = true;
                         }
                         else
                         {
-                            web.AssociatedOwnerGroup = EnsureGroup(web, parsedAssociatedOwnerGroupName);
-                        }
+                            if (!web.GroupExists(parsedAssociatedOwnerGroupName))
+                            {
+                                web.AssociatedOwnerGroup = EnsureGroup(web, parsedAssociatedOwnerGroupName);
+                                webNeedsUpdate = true;
+                            }
+                            else if (!web.AssociatedOwnerGroup.ServerObjectIsNull.Value)
+                            {
+                                if (web.AssociatedOwnerGroup.Title != parsedAssociatedOwnerGroupName)
+                                {
+                                    var updatedOwnerGroup = web.SiteGroups.GetByName(parsedAssociatedOwnerGroupName);
+                                    web.Context.Load(updatedOwnerGroup);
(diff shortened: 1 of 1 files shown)


heinrich-ulbricht:
Wouldn't your code still create a new associated group if the group name in the template differs from any group name in the destination site? Or am I reading this wrong?
```
if (!web.GroupExists(parsedAssociatedOwnerGroupName))
{
    web.AssociatedOwnerGroup = EnsureGroup(web, parsedAssociatedOwnerGroupName);
    webNeedsUpdate = true;
}
```

gautamdsheth:
You are correct. We also have to consider a scenario where this is an intended behaviour. If for example you don't want to create associated groups, then you can simply modify the XML from:

`<pnp:Security RemoveExistingUniqueRoleAssignments="false" AssociatedOwnerGroup="Contoso Electronics Owners" AssociatedMemberGroup="Contoso Electronics Members" AssociatedVisitorGroup="Contoso Electronics Visitors">`

to this

`<pnp:Security RemoveExistingUniqueRoleAssignments="false">`

heinrich-ulbricht:
I agree that we shouldn't block legitimate scenarios.
But I'd consider #2127 still open in this case. The ticket is explicitly about preventing 2 co-existing owner/member/visitor groups (one of which is not the "real one") which seems confusing to me as default behavior.
Maybe we can come up with a simple solution to make the default behavior more intuitive while still being able to force the creation if required (all without having to dive into the XML which the casual user won't do).

heinrich-ulbricht:
@gautamdsheth Just out of cursiosity: what's the background of checking for noscript sites?
https://github.com/SharePoint/PnP-Sites-Core/blob/30f53bd487a73e53b4d074992e1f89d87885f6b5/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs#L44

Is there a difference regarding associated groups? In #2127 I used a modern team site for testing - maybe I should add this information to the ticket?

gautamdsheth:
@heinrich-ulbricht  This basically checks the site collection's DenyAddAndCustomizePages property. This is enabled on modern sites. You can get more details [here](https://docs.microsoft.com/en-in/sharepoint/allow-or-prevent-custom-script).

jansenbe:
@gautamdsheth : did you see issues while testing this PR on sites that have NoScript enabled?

jansenbe:
Merged, but removed the check on NoScript sites

gautamdsheth:
While executing this on Modern sites,new  groups get provisioned but are not set as associated groups. I got access denied error while setting the new groups. That is the reason while I made the NoScript check, can you check ?  This only happens on modern sites, on classic sites it works fine.

jansenbe:
hmm...that's due to the async nature of the provisioning...I recall having seen other issues due to that. I'll add the check again

jensotto:
@jansenbe I don't understand why this PR has been approved. The test method CanProvisionNewAssociatedGroup fails with this new code. Have you not seen this?

jensotto:
@jansenbe I see now why the code fails. This is due to a unwanted side effect when using the PnP extension SecurityExtensions.GroupExists(string). The call to web.SiteGroups.GetByName(string) adds the group name to the locally cahced list of groups. This is not cleared if the group is not found and a server exception is returned. Later calls to the same method (SiteGroups.GetByName(string)) will then incorrectly assume that the group exists on the server. This results in that the code in ObjectSiteSecurity.EnsureGroup(Web, string) never creates the group on the server.
The conclusion is therefore that this PR needs to be reverted, and SecurityExtensions.GroupExists(string) needs to be rewritten to use an ExceptionHandlingScope.
I can probably contribute this, but I'm not sure when I will be able to do this.

jensotto:
@jansenbe I also fail to see how this PR actually fixes #2127
Due to the code failing as previously mentioned, it will not complete a successful run in the situation mentioned in #2127

jensotto:
This PR also fails the unit test ObjectSiteSecurityTests.CanProvisionAssociatedGroupsInSubSite() in all version of SharePoint.

PR #2136, opened 2019-03-27
EnsureUser for external users

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | N/A

Earlier it wasn't possible to EnsureUser for an extern user.
However, now if an external user exist in directory for the tenant, it is possible to EnsureUser on a Site where the user doesn't exist within the Site Collection.

So instead of skipping external users this change will allow the provisioning engine to try to EnsureUser and log a Warning if it didn't succeed.


jansenbe:
Thanks @SchauDK

PR #2121, opened 2019-03-20
Feature: Added support for OOTB SPO themes via ThemeManager extension

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | NA

#### What's in this Pull Request?

This PR adds additional support for applying SharePoint Online's default available themes to a given web. 

While we are able to apply our own custom themes via Theme Manager, there was no such way to apply OOTB default available themes to a SP web.

PR #2142, opened 2019-03-29
Apply out of the box modern theme

anoopt:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no 
| Related issues?  | NA

#### New bits
This PR has a new extension method called "ApplyOOBModernTheme" which can be used to apply an out of the box modern theme to a site. 
The JSON for theme is created by a new method called "GetOOBModernThemeJson" in "ThemeUtility.cs". 
There is also an enum called "OOBTheme" which has all the out of the box theme names.

#### Existing bits
Thanks to Paolo, the already existing "ApplyThemeAsync" is used for calling the "ApplyTheme" endpoint. 

Thanks to Gautam Sheth for the PR suggestion.


Core/OfficeDevPnP.Core/Utilities/Themes/ThemeUtility.cs
```diff
@@ -48,6 +48,36 @@ public static String GetThemeAsJSON(String primaryColor, String bodyTextColor, S
...
         }
+
+        /// <summary>
+        /// Return an out of the box modern theme as a JSON string
+        /// </summary>
+        /// <param name="theme">Name of the out of the box theme</param>
+        /// <returns></returns>
+        public static String GetOOBModernThemeJson(OOBTheme theme)
+        {
+            switch (theme)
```
heinrich-ulbricht: Interesting. I just had a look at the "Change the Look" options and the theme colors. They seem to be embedded in the JS bundle. So no API to get the colors from (which I expected). But how do we detect if the default colors change? (Does it even matter?)

anoopt: True. Even I did not see any API which provides the OOB theme information. So one option is to use the code in the PR and as soon as there is one I will update the code.

erwinvanhunen:
Hi! Thank you for the PR. However, we merged another PR (#2121) which does exactly the same, in a somewhat different (but similar) implementation. Unfortunately this means we will not merge this PR. Nevertheless, thank you very much for your efforts! We really appreciate your work.

anoopt:
Ah. I didn't know there was another PR for this. My bad. Apologies.

PR #2150, opened 2019-04-05
Allow fileuniqueid in NavigationNodes for Group OneNote Url

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | Group Default OneNote Link uses fileuniqueid in Navigation Url

**What's in this Pull Request?**
In ObjectNavigation added Function ReplaceFileUniqueToken to replace fileuniqueid or fileuniqueidencoded if exist in Navigation Url.

The reason for this is since the Group Note Note Link uses the fileuniqueid in the Url and if you want to create the Link by Template you will need to be able to replace the token pointing to the SiteAssets One Note Folder.

@jansenbe Clean re-checkin with single change - sorry for pull before


jansenbe:
Thanks @czullu

PR #2153, opened 2019-04-05
Fix of bug "Taxonomy import creates all terms in the first term set s…

### LINKED ISSUE #581: Taxonomy import creates all terms in the first term set specified in the input file
#### Category
[ X] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ X] SharePoint 2013

If SharePoint on-premises, what's exact CU version: September 2015 CU

The problem is related to exporting and importing taxonomy while using PnP-PowerShell module.  I would expect to have the same hierarchy to be applied after import as it existed from the export. 
 
#### Observed Behavior
What ends up happening is that the term groups all get created, but only the first term set in the import file is created and all terms are imported under this single term set instead of the expected hierarchy.

#### Steps to Reproduce
I have created term groups, term sets, and terms in the managed metadata service application manually and used the Export-SPOTaxonomy cmdlet.  I then use the Import-SPOTaxonomy to import the terms.

IonutLupsan:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #581 

The ImportTerms method of the Taxonomy extensions now groups the input lines also by termsets, not only by term groups. The following ImportTermset method is then called for each termset in each term group.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs
@@ -569,7 +569,7 @@ namespace Microsoft.SharePoint.Client
         /// <param name="synchronizeDeletions">Remove tags that are not present in the import</param>
         public static void ImportTerms(this Site site, string[] termLines, int lcid, TermStore termStore, string delimiter = "|", bool synchronizeDeletions = false)
         {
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you @IonutLupsan! Apologies for the long wait. Great fix!

PR #2159, opened 2019-04-09
PageExport of data-sp-htmlproperties to Template and Provisioning to Page fixed

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

ExtractClientSidePage exports now JSON Encoded the HtmlProperties and HtmlPropertiesData from ClientSideWebPart to JsonControlData of Control so they will be in PnP-Template.xml 

RenderHtmlProperties of ClientSideWebPart gives HtmlPropertiesData if they exist

@jansenbe : Fix HtmlProperties and HtmlPropertiesData on Export & Provision since we had issues with WebParts missing this data after provisioning the Page. To my understanding the data structure in this fields is up to the Programmer so i just encode content with JsonConvert.ToString(HtmlPropertiesData, '"', StringEscapeHandling.EscapeHtml) in order not to break the rest of the JSON Object.


jansenbe:
Hi @czullu ,

I'm not sure why you need to export these, especially if serverProcessedContent is available? ServerProcessedContent is the JSON representation of HtmlPropertiesData and at save time we translate ServerProcessedContent into the correct HtmlPropertiesData structure. 

Are you trying to fix the case where that there's no ServerProcessedContent data available but there is the HtmlPropertiesData?

czullu:
Hi Bert

Currently I'm on vacation with limited access to the computer.

What I recall, I have the case where the webpart does not show content until I edit the page and save it again after provision page. If I do edit and save the hmtlpropertiesdata is restored. In order to have content shown after provision I had to restore htmlpropertiesdata as well.

I will see that I get access tomorrow and follow up also your other questions.

Best regards
Christian

Freundliche Grüsse
Christian

jansenbe:
No rush, enjoy your vacation!

jansenbe:
Hi @czullu ,

I'm closing this PR as I we can't merge it. If that additional data needs to be exported then we need to somehow translate it into serverProcessedContent I think. Would also need a repro of the problem so that we can verify the solution.

Issue #2160, opened 2019-04-10
Modern List "Add Column" creates Issue on extract/provision template


czullu:
#### Which PnP repository should you use to report the issue? 
- PnP Sites Core - https://github.com/OfficeDev/PnP-Sites-Core

#### Category
[x] Bug
[ ] Enhancement

#### Environment
[x] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

#### Expected or Desired Behavior
The User creates a new Column on for example a Document Library with the "Add Column" button (it does not matter if he choose all contenttypes or specific under Options). 
Extract Template and reapply to a new site. The expectation would be that this column now exists on the List in the New Site and can be used.

#### Observed Behavior
The new Column gets Added to the List but is not assigned to any ContentType of the List and therefore not usable. If it happens that you use ListDataRow ->FieldValues the Provisioning will fail with the Message that the Column is Read-Only ( a bit missleading - but still somehow correct).

Take a Look to XML we find the Custom Field in <pnp:ListInstance><pnp:Fields> Section so thats why it's created. 
My Idea to just add the Field to the <pnp:FieldRefs> of the List fails because the Code will not find a Field defined just on List Level. 
For this i will provide a fix - but it will have the downside of adding the Field to all Content Type on the List and can be argued.

The clean way would be to extend the Schema and allow under  <pnp:ListInstance><pnp:ContentTypeBindings><pnp:ContentTypeBinding> new some <pnp:FieldRefs> in order to tell to which ContentType you want to add this Column.

#### Steps to Reproduce
Use "Add Column" on modern List, extract to Template and provision to new site. since you will have the new Column by default in the View you will find it empty and not shown for edit / not assigned to any Content Type on List. 

@PaoloPia @jansenbe : Im open for discussion whats the best way to fix List Level Fields Issue but since it got so easy for users to Add Columns in Modern Lists i'm afraid that this Issue will pop-up more often. I mention you both because of Schema & Provision Engine

PR #2163, opened 2019-04-11
Fix List Level column not added to ContentType of List

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | #2160

While the ProcessFieldRefs did add List Level Fields to Default ContentType if just one ContentType was on the List it did not add the Field to any ContentType if there where multiple ContentTypes on the List.

The Logic is now Changed in the way that if it's a FieldRef to a Field defined on the List Level of the Template and the Field is not already Linked to any of the ContentTypes on the List i will Link the Field to any of the existing ContentTypes on the List. 

This will fix the Issue restoring FieldValues to such a List Column added by Modern List "Add Column" as well as making the Column not only appear in ListView but also be editable.

As it is not perfect, it's the best i could think of to handle this Issue


PaoloPia:
Hi @czullu,
Thanks for your contribution. I cannot merge this PR, because there are conflicts with the actual code. However, I applied the same changes to the current version of the code. You will be still credited for your contribution.

PR #2164, opened 2019-04-11
Fix : Compile issues for OnPrem in the Site Header/Footer handlers

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes NA

1) This PR fixes compile issues while building the solution for all OnPrem environments
2) Used the helper method `web.GetBaseTemplateId();`  in both Footer and Site Header handlers
3) `ExecuteQueryRetry()` was missing in the Site Header handler, so added that 
4) Also added support for Site headers in Modern team site (STS#3) templates as they are supported there as well. 

To be fair, the site headers are supported in the Modern UI of even classic site collections, so can we remove this check and simply extract it for all templates ? Let me know your thoughts on this, if needed can modify my PR :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Support for provisioning Site Footer links 
 - New PnPProvisioningContext object for security scope management
 - New tenant extension method (tenant.EnableCommSite) to to convert the root site collection of a tenant into a communication site
+- Added support to extract and provision list propertybag entries #2201 [patrikhellgren]
 
 ### Changed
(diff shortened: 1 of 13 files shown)


erwinvanhunen:
Hi! Can I ask you to revisit this PR and handle the conflicts? We've updated the engine in the mean time and we filtered the handlers for on-prem a bit differently. Both ways are valid as such I think. And regarding the Header, yes, you can set that property when on a modern page in a classic site. We can remove that test indeed.

gautamdsheth:
Closing this PR as i made some mistake merge conflict. Sending a new PR.

PR #2166, opened 2019-04-11
Add support for TermGroups and TermSets scoped on SiteCollection

freekbe:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no

Support for Termgroups and Termsets that are created on SiteCollection level.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectHandlerBase.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectHandlerBase.cs
@@ -110,7 +110,14 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                     {
                         termSet.EnsureProperties(ts => ts.Name, ts => ts.Group);
 
-                        termSetIdElement.Value = String.Format("{{termsetid:{0}:{1}}}", termSet.Group.IsSiteCollectionGroup ? "{sitecollectiontermgroupname}" : termSet.Group.Name, termSet.Name);
+                        if (termSet.Group.IsSiteCollectionGroup)
+                        {
+                            termSetIdElement.Value = String.Format("{{sitecollectiontermsetid:{0}}}", termSet.Name);
+                        }
+                        else
+                        {
+                            termSetIdElement.Value = String.Format("{{termsetid:{0}:{1}}}", termSet.Group.Name, termSet.Name);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


erwinvanhunen:
Hi! I don't follow exactly what you are trying accomplish in your code? The TermSetIdToken that is being added to the token cache in the parser requires the name of the actual termset (and not the token label as you did) and the id of the termset. Adding the token as you are doing in the tokenparser.cs would have no effect as such:

the token {sitecollectiontermgroupname} token is already being added to the tokenparser cache as such, so it should resolve if being used. We support nested tokens, so things should just work as such given that the token is already in the cache. Notice also that tokens are resolved as much as possible at runtime. Checkout the SiteCollectionTermGroupName class which is available in the TokenDefinitions folder of the engine.

Could you elaborate a bit on the specific scenario you are trying to solve here, to make sure I'm not misunderstanding this PR?

freekbe:
Hi Erwin,

There is an issue with the tokens for TermSets that are scoped on
SiteCollection level. The SiteCollection TermGroup is created successfully
but then the creation of the TermSet fails because the token cannot be
resolved.
Adding the token to the cache (as I did in the TokenParser.cs) solves this
issue.

I don't have access to my test environment at this time but if I remember
correctly, the token that was added to the cache was using a wrong name for
the TermGroup. Replacing the TermGroup name with the token
{sitecollectiontermgroupname} solves this problem because it gets replaced
by the correct value at runtime, as you mention.

Let me know if this is not clear, I can send you more details once I have
access again to my test environment.

Op ma 29 apr. 2019 om 15:16 schreef Erwin van Hunen <
notifications@github.com>:

erwinvanhunen:
Hi Freek, please do. If the token is not being added correctly or does not resolve correct then we need to fix that in the token parser at that part (lhttps://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs#L362) or at https://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenDefinitions/SiteCollectionTermGroupNameToken.cs

freekbe:
Hi Erwin,

I updated this PR to make use of "SiteCollectionTermSetIdToken".
When using this token, there  was an issue to link taxonomy fields to the correct SiteColletion TermSet. The token that was defined during the extraction of a taxonomy field was not correct. This is fixed now as well.

freekbe:
One more thing: If this PR is accepted, I'm not sure if this block of code is still needed: https://github.com/SharePoint/PnP-Sites-Core/blob/d5f95c7189bd5b96fa2bb1690c95195b2215ed6d/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs#L358-L388
That code is also creating the necessary SiteCollectionTermSetIdTokens but "only when we're not working in app-only".

jansenbe:
Hi,

This PR has been open for a long time and we apologize for not having been able to timely process it. Given things have changed a lot over the last year this PR needs to be revisited and updated if the requested change is still relevant. If still relevant then please create a new PR, verified against our dev branch. In the meanwhile we'll be closing this PR, hope you understand.

PR #2174, opened 2019-04-16
Replace associated group title with site title token during extraction

### LINKED ISSUE #2127: Associated groups from template should not be created

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #2127, reverts #2128

Revert #2127 as it broke the functionality in SP2013 and blocked changing associated groups if the groups already existed in the site

Modifies the extraction of associated owner, member, and visitor groups to replace group title with site title token.
Adds option to include extraction of associated owner, member, and visitor groups. (IncludeDefaultAssociatedGroups)
I could not find a better naming for this property as I did not want to use IncludeAssociatedGroups. That name could in the future collide with the actual Associated Groups which is not the same as the associated owner, member, and visitor groups.
I'm open for other suggestions for naming this property differently.

Improved code checking for existence of a group to work properly on SharePoint 2013.

Added utility method to run code with DisableReturnValueCache on the specified ClientRuntimeContext object.
Setting DisableReturnValueCache in ClientContextExtensions.ExecuteQueryImplementation should not be done as this creates an unwanted side effect when for example using ExecuteQueryRetry, especially when using it outside the OfficeDevPnP.Core framework.
In addition setting the value of DisableReturnValueCache right before calling ExecuteQuery has no effect. It needs to be set prior to using any of the methods that caches the return value. I'll create a separate bug for discussing this further (#2175).

#### Important:
I'm unable to run tests for any of the on premises versions (debug15, debug16, and debug19) because of many recent changes that has made the code not build in those configurations anymore.

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
@@ -269,6 +269,7 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
                 // Load the base template which will be used for the comparison work
                 var creationInfo = new ProvisioningTemplateCreationInformation(web) { BaseTemplate = web.GetBaseTemplate() };
                 creationInfo.IncludeSiteGroups = true;
+                creationInfo.IncludeDefaultAssociatedGroups = true;
                 var template = new ProvisioningTemplate();
                 template = new ObjectSiteSecurity().ExtractObjects(web, template, creationInfo);
 
@@ -305,10 +306,23 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
                 var template = new ProvisioningTemplate();
                 template = new ObjectSiteSecurity().ExtractObjects(web, template, creationInfo);
 
+                Assert.IsFalse(template.Security.SiteGroups.Any());
+            }
+        }
+
+        public void CanSkipExtractAssociatedGroups()
+        {
+            using (var clientContext = TestCommon.CreateClientContext())
+            {
+                Web web = clientContext.Web;
+                var creationInfo = new ProvisioningTemplateCreationInformation(web) { BaseTemplate = web.GetBaseTemplate() };
+                creationInfo.IncludeDefaultAssociatedGroups = false;
+                var template = new ProvisioningTemplate();
+                template = new ObjectSiteSecurity().ExtractObjects(web, template, creationInfo);
+
                 Assert.IsNull(template.Security.AssociatedOwnerGroup);
                 Assert.IsNull(template.Security.AssociatedMemberGroup);
                 Assert.IsNull(template.Security.AssociatedVisitorGroup);
-                Assert.IsFalse(template.Security.SiteGroups.Any());
             }
         }
 
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
@@ -249,13 +249,6 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
         {
             using (var ctx = TestCommon.CreateClientContext())
             {
-                Web web = ctx.Web;
-                ctx.Load(web,
-                    w => w.AssociatedOwnerGroup.Title,
-                    w => w.AssociatedMemberGroup.Title,
-                    w => w.AssociatedVisitorGroup.Title);
-                ctx.ExecuteQuery();
-
                 // Load the base template which will be used for the comparison work
                 var creationInfo = new ProvisioningTemplateCreationInformation(ctx.Web) { BaseTemplate = ctx.Web.GetBaseTemplate() };
 
@@ -263,9 +256,6 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
                 template = new ObjectSiteSecurity().ExtractObjects(ctx.Web, template, creationInfo);
 
                 Assert.IsTrue(template.Security.AdditionalAdministrators.Any());
-                Assert.AreEqual(web.AssociatedOwnerGroup.Title, template.Security.AssociatedOwnerGroup);
-                Assert.AreEqual(web.AssociatedMemberGroup.Title, template.Security.AssociatedMemberGroup);
-                Assert.AreEqual(web.AssociatedVisitorGroup.Title, template.Security.AssociatedVisitorGroup);
             }
         }
 
@@ -274,15 +264,51 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
         {
             using (var ctx = TestCommon.CreateClientContext())
             {
+                Web web = ctx.Web;
+
                 // Load the base template which will be used for the comparison work
-                var creationInfo = new ProvisioningTemplateCreationInformation(ctx.Web) { BaseTemplate = ctx.Web.GetBaseTemplate() };
+                var creationInfo = new ProvisioningTemplateCreationInformation(web) { BaseTemplate = web.GetBaseTemplate() };
                 creationInfo.IncludeSiteGroups = true;
                 var template = new ProvisioningTemplate();
-                template = new ObjectSiteSecurity().ExtractObjects(ctx.Web, template, creationInfo);
+                template = new ObjectSiteSecurity().ExtractObjects(web, template, creationInfo);
+
+                ctx.Load(web,
+                    w => w.AssociatedOwnerGroup.Title,
+                    w => w.AssociatedMemberGroup.Title,
+                    w => w.AssociatedVisitorGroup.Title);
+                ctx.ExecuteQuery();
 
                 Assert.IsTrue(template.Security.AdditionalAdministrators.Any());
                 Assert.IsTrue(template.Security.SiteGroups.Any());
+                Assert.AreEqual(SiteTitleToken.GetReplaceToken(web.AssociatedOwnerGroup.Title, web), template.Security.AssociatedOwnerGroup, "Associated owner group title mismatch.");
+                Assert.AreEqual(SiteTitleToken.GetReplaceToken(web.AssociatedMemberGroup.Title, web), template.Security.AssociatedMemberGroup, "Associated member group title mismatch.");
+                Assert.AreEqual(SiteTitleToken.GetReplaceToken(web.AssociatedVisitorGroup.Title, web), template.Security.AssociatedVisitorGroup, "Associated visitor group title mismatch.");
 
+                // These three assertions will fail if the site collection does not have the
+                // default groups created during site creation assigned as associated owner group,
+                // associated member group, and associated visitor group.
+                // This is a prerequisite for the site collection used for unit testing purposes.                
+                Assert.IsTrue(template.Security.AssociatedOwnerGroup.Contains("{sitetitle}"), "Associated owner group title does not contain the Site Title token.");
+                Assert.IsTrue(template.Security.AssociatedMemberGroup.Contains("{sitetitle}"), "Associated owner group title does not contain the Site Title token.");
+                Assert.IsTrue(template.Security.AssociatedVisitorGroup.Contains("{sitetitle}"), "Associated owner group title does not contain the Site Title token.");
+            }
+        }
+
+        [TestMethod]
+        public void CanSkipExtractSiteGroups()
+        {
+            using (var clientContext = TestCommon.CreateClientContext())
+            {
+                Web web = clientContext.Web;
+                var creationInfo = new ProvisioningTemplateCreationInformation(web) { BaseTemplate = web.GetBaseTemplate() };
+                creationInfo.IncludeSiteGroups = false;
+                var template = new ProvisioningTemplate();
+                template = new ObjectSiteSecurity().ExtractObjects(web, template, creationInfo);
+
+                Assert.IsNull(template.Security.AssociatedOwnerGroup);
+                Assert.IsNull(template.Security.AssociatedMemberGroup);
+                Assert.IsNull(template.Security.AssociatedVisitorGroup);
+                Assert.IsFalse(template.Security.SiteGroups.Any());
             }
         }
 
--- a/Core/OfficeDevPnP.Core/Extensions/SecurityExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/SecurityExtensions.cs
@@ -1134,6 +1134,16 @@ namespace Microsoft.SharePoint.Client
         /// <returns>True if the group exists, false otherwise</returns>
         public static bool GroupExists(this Web web, string groupName)
         {
+            if (web == null)
+            {
+                throw new ArgumentNullException(nameof(web));
+            }
+
+            if (string.IsNullOrEmpty(groupName))
+            {
+                throw new ArgumentNullException(nameof(groupName));
+            }
+
 #if SP2013
             IEnumerable<Group> groups;
             groups = web.Context.LoadQuery(web.SiteGroups.Include(g => g.LoginName));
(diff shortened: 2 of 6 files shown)


jensotto:
@gautamdsheth Can you please verify that this change works for your requirements? Your changes in #2128 broke the functionality in SP2013 and blocked changing associated groups if the groups already existed in the site.

gautamdsheth:
@jensotto - Firstly, apologies for breaking the stuff. That was certainly not my intention :( 
Let me check this around the weekend and get back to you on this. Will do that first thing around this weekend. 

One question -  you mentioned in my PR that, " `web.SiteGroups.GetByName(string)` adds the group name to the locally cached list of groups" , however i couldn't reproduce that behaviour in SPO ? How is that possible, do you have any idea ?

Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
```diff
@@ -274,15 +264,51 @@ public void CanCreateEntities2()
...
+
+                ctx.Load(web,
+                    w => w.AssociatedOwnerGroup.Title,
+                    w => w.AssociatedMemberGroup.Title,
+                    w => w.AssociatedVisitorGroup.Title);
+                ctx.ExecuteQuery();
 
                 Assert.IsTrue(template.Security.AdditionalAdministrators.Any());
                 Assert.IsTrue(template.Security.SiteGroups.Any());
+                Assert.AreEqual(SiteTitleToken.GetReplaceToken(web.AssociatedOwnerGroup.Title, web), template.Security.AssociatedOwnerGroup, "Associated owner group title mismatch.");
```
heinrich-ulbricht: @jensotto Could this fail when the site used for testing is created right before this test? There is some asynchronous action going on that initially names the default associated groups like "Site Members" (without the title) and then some seconds (minutes?) after site creation changes the group names to include the site title.
Just want to be sure not to introduce timing issues here making the test fail sometimes and sometimes not.

jensotto: That should not matter as this is only a unit test. The unit tests runs against an already existing site collection, so this should not be an issue.

heinrich-ulbricht: Ok I'll try another angle. I'm concerned about code like the following and timing issues arising from the fact that SharePoint changes the group name behind our back several seconds after running this code:

https://github.com/SharePoint/PnP-Sites-Core/blob/93295420e78549a496936d71ec07655255fda2e0/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs#L64

This could create a group although it is not necessary.

I just tested with a newly created site. On my developer tenant it takes 18 seconds (!) after site creation for the initial group names to change from e.g. "Site Owners" to "Site Title Owners". So code checking group names has a time window of 18 seconds where it sees the "old" names. And then suddenly the "new" names.

(Somewhere I saw a commend I think by @jansenbe referring to timing issues and site creation.)

My thesis is that we have to wait for the group names to be "final" before doing anything with them. So this would be an 18-second wait in my specific case. And my questions are: are my concerns reasonable? If yes: is this already being taken care about in the framework? Maybe @VesaJuvonen could chime in and give us a hint - do we have to care about this delay of several seconds between site creation and changing the group names?

gautamdsheth:
@jensotto - The changes look good. I have made a few additions to it, mainly to avoid extra processing /check of already present groups. This will be in case we apply the same template on the same site/web and we dont make any modifications to the group names. Can you please review it ? I couldn't send a PR to your repo, so upload the changes here itself. 


[ObjectSiteSecurity.txt](https://github.com/SharePoint/PnP-Sites-Core/files/3100126/ObjectSiteSecurity.txt)

Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
heinrich-ulbricht: You know what? I created a test that checks what I'd expect. No need to speculate :) -> https://github.com/SharePoint/PnP-Sites-Core/pull/2192

jensotto:
@gautamdsheth I understand that it could be cleaner to skip setting the associated groups when the same ones are already set. There is however one issue with how we can identify if the group is the same as the one specified in the template. Internally SharePoint uses a culture specific, case insensitive comparison for comparing the group title. It's easy to provide a case insensitive comparison client side, but the culture specific comparison information is not currently possible to do exactly as done internally in SharePoint. SharePoint uses the web locale to create the culture specific comparison information, and this is not available client side. In order to make some of the code work with SharePoint 2013 I had to use the web language for culture specific comparison information. This is not the same as what SharePoint internally does, and it could therefore lead to incorrect comparison in some extreme cases. I therefore think it's best to avoid such comparisons client side. The downside of this is that we have to assign the group to the associated owner/member/visitor group property and call Web.Update() every time the template is applied. 
There is however one approach that would work, and that is to compare the group id, but I think that would be counterproductive as it requires an extra roundtrip to the server to fetch the group id after calling ObjectSiteSecurity.EnsureGroup(string).
I therefore think it's better to leave this as it is for now, until Microsoft exposes Web.Locale to CSOM.

heinrich-ulbricht:
@jensotto Do you know of a ticket somewhere ([UserVoice](https://sharepoint.uservoice.com/forums/329220-sharepoint-dev-platform?category_id=194593)?) for exposing SPWeb.Locale to CSOM?

jensotto:
@heinrich-ulbricht No I don't. Maybe we should just create one.

jansenbe:
Hi @jensotto , @heinrich-ulbricht , @gautamdsheth ,

I've been looking into this PR and the issue it's solving and have some feedback/questions:
- Using the SiteTitle token makes a lot of sense: I've changed that

About group creation issues (duplicate groups vs being able to change the associated groups) we indeed have some challenges due to the async provisioning of modern sites. Matching a group name is not reliable anymore as you've found out. What do you think about the below high-level flow:

Check is associated groups are set?
  - **No, groups are not set**: create the groups according to what's defined inside the template (this should only apply to on-premises as online sites get associated groups by default)
 - **Yes, groups are set**: Resolve group names and check if that group name is listed an additional group in the security section of the template
    - **Not listed as a group in the template**: in this case, we assume the default associated groups should be used...we're not checking for their name as this is not reliable with async provisioning...we checked previously that there was an associated group so we're good
    - **Listed as a group in the template**: this is a purposely defined replacement of an associated group so let's change the associated group

heinrich-ulbricht:
@jansenbe I want to make sure I understand this correctly.

You are proposing to check two things in the provisioning template: 

1. presence of the `Associated*Group` attributes on the `Security` element and 
2. the presence of `SiteGroup` elements with the _same_ group names.

Let's number your cases (I added a third to be clear about what "_groups are set_" means):

1. No, groups are not set (that means: no associated groups are set in the site)
2. Resolve group names and check if that group name is listed as an additional group in the security section of the template
   1. Not listed as a group in the template
   2. Listed as a group in the template
3. No associated `Associated*Group` attributes are defined in the template

**So case 2.1. looks like this:**

```
      <pnp:Security RemoveExistingUniqueRoleAssignments="false" AssociatedOwnerGroup="Site Owners">
        <pnp:SiteGroups>
        </pnp:SiteGroups>
      </pnp:Security>

```
The Attribute `AssociatedOwnerGroups="Site Owners"` will be ignored in this case, additional members for this group will be added to the already existing associated group of the site. Is this correct?

**Case 2.2. looks like this:**

```
      <pnp:Security RemoveExistingUniqueRoleAssignments="false" AssociatedOwnerGroup="Site Owners">
        <pnp:SiteGroups>
          <pnp:SiteGroup Title="Site Owners" />
        </pnp:SiteGroups>
      </pnp:Security>

```
Since the group is explicitly listed as `SiteGroup` element it will be created and assigned as new `AssociatedOwnerGroup`. 
Edge cases: 

- When after resolving a site title token the resulting name collides with the name of the existing associated owner group name the existing associated group will be used. Yes/no?
- What if the existing group with same name is no associated group yet? Should this existing group be promoted to an associated group?

**Case 3. might look like this:**

```
      <pnp:Security RemoveExistingUniqueRoleAssignments="false">
        <pnp:SiteGroups>
          <pnp:SiteGroup Title="Site Owners" [...]>
        </pnp:SiteGroups>
      </pnp:Security>

```

No changes to associated groups will be done in this case.

Did I understand this correctly?

jansenbe:
Case 2.1 ==> Yes, correct
Case 2.2 ==> If the group name is found, then yes let's reuse that group (this allows a template to be applied multiple times). If the existing group is not an associated group then let's make it the associated group
Case 3 ==> Correct

Thanks for thinking along @heinrich-ulbricht...above is a proposal in order to get a final fix that will be OK for everyone. As we plan to push out the May release soon I would like to get this handled in the coming days...if @jensotto and @gautamdsheth  are ok with the above then someone can implement it (alternatively I'll take a shot at it).

heinrich-ulbricht:
This someone could as well be me but I'd also like to hear feedback from @jensotto and @gautamdsheth since they have a strong opinion about this case. From my perspective the most efficient would be if one of you guys found the time to modify the code since you already wrapped your head around the details.  :man_technologist:

jensotto:
@jansenbe Does this solve the async provisioning issue in SharePoint Online? It seems to me that templates with both an associated group and identically named site group specified will still run into issues with group names being changed after a few minutes. This assumes that the site title token is used in the group title.
I also see a few issues with the group creation process from the template. Default values are imposed on groups even if properties are not specified on the template. It appears that groups are skipped if the title matches the owner, or if it contains the one of the associated members token. The point is that there might be issues if we force the template to specify associated groups in the site group templates without supporting nullable values for the site group properties and also supporting all different values of owners.
It's too late for me to dive into this now, and I'm afraid I don't have time to look further into this this week.

heinrich-ulbricht:
Hm. What now? @jansenbe Would you say "Go" and refine later or wait and discuss this through? In any case the implementation would have to be accompanied by tests describing the desired outcome.

jansenbe:
@jensotto : In my view, this does solve the async provisioning issue as we're not doing any name based lookups anymore, so using the sitetitle token or not is not making a difference. I've also updated the extraction logic to use associatedgroupid tokens which work in a name independent way.

The other concerns you mention have been there since a while, so let's handle them separately...first focus should be to ensure the key scenarios work (no duplicate group when async provisioning is used and the ability to correctly set associated groups for both on-premises as cloud versions).

@heinrich-ulbricht : if you've some bandwidth for this that would be great...you could then as part of this update your pending PR with the associated modern group provisioning test case. If there's no time than that obviously is a valid answer as well. From a timing perspective, we'll push the May master merge to end of next week, so there's some more time.

heinrich-ulbricht:
I'll have a look at the code. @jansenbe Which branches should I take into account? Just dev? You mentioned that you updated something regarding associatedgroupid (in dev?) - does this cover logic from https://github.com/jensotto/PnP-Sites-Core/tree/associated-groups-token? Do I need https://github.com/jensotto/PnP-Sites-Core/tree/associated-groups-token or is this obsolete due to changes you did?

jansenbe:
My change is in the dev branch already. I did not check Jens branch but this id based token has been in use for a couple of months already. I just changed the extraction logic to favor the id based token over the name based token

heinrich-ulbricht:
I got some questions while implementing this.

Does the `{sitetitle}` token replacement still make sense?
Not for case 2.1 since we want to ignore the name of the associated groups. (Side question: Would a * or some other placeholder be a more user friendly value in the template?)
And for case 2.2? Are the group names tokenized as well, or just the `Associated*Group` attributes? Not sure about this, but they should be, right?.

I updated my pull request to show the current state for review. I rolled back the changes by @gautamdsheth and modified the logic according to the points discussed above. The tests from @jensotto are broken as the assumptions are not true anymore. They also contain checks for tokens and I don't want to touch this until the questions above are answered.
Also, the new check for NoScriptSites introduced by @jansenbe / @gautamdsheth makes those tests break even more since the whole Associated*Group magic is skipped in this case. I mitigated this by temporarily making the test sites script-enabled.

And a closing thought: with our group name mapping we potentially live in two worlds: one before the async group rename by SharePoint and one after the async group rename. Can we determine which "world" the template is being applied? Can we detect that a group is the associated group before the async rename takes place? (By checking the ID for a well known value maybe?) IF we can detect this state we could modify the mapping behavior at least for point 2.1 to not completely ignore the group names but to interpret the `{sitetitle}` token as matching both incarnations of the default associated group name - eg. "Site Owners" _and_ "Site Title Owners", but nothing else.

I'm not completely satisfied that we ignore values in the template file in case 2.1 and also when applying a template to a NoScript site. This is confusing from the user's standpoint.

I hope this makes sense.

What are the next steps? This depends on your feedback. And I'd like to add more tests to also cover NoScript sites since the current implementation switches off the NoScript flag.

gautamdsheth:
I would suggest that we modify the `noScript` check and instead check if the underlying web Template is communication site or Team site (O365 group). If they are communication/team sites, we skip this entire logic of associated groups for these 2 web templates.

They are kinda special sites and changing the associated groups might have some issues later on. For example, in the modern team sites(O365 group), it can impact other O365 artefacts as well like Teams or Stream. Also, we got access denied while provisioning new associated groups in these sites. Did you facing something similar as well while provisioning for modern sites ? 

Fixing them by flipping the NoScript flag seems a bit too heavy IMHO and would require tenant level permissions. Not sure everybody will have it and it does look like a breaking change. We can write a warning message for the users saying something like "New associated groups wont be created for modern sites like team or communication site, we will be reusing the already existing".  Thoughts ? 

To determine the templates of modern sites, you can use the below code:

```
var webTemplate = web.GetBaseTemplateId();
if(webTemplate == "SITEPAGEPUBLISHING#0" || webTemplate == "GROUP#0")
{
//skip creating associated groups
}
else 
{
// process the changes in associated groups
}

```

Thoughts ?

heinrich-ulbricht:
To be clear: I'm **not** proposing switching the NoScript flag off for modern sites as part of the provisioning. I'm doing it **only** in the tests to be able to test the behavior for script-enabled sites. I was just too lazy to change the communication test sites I use to something else.
Modern sites seem to be NoScript sites so I think the check for the template is not necessary. I like the idea of the warning though, so the user instantly sees that some of his Associated*Groups configuration won't be applied.

heinrich-ulbricht:
I'm adding the warning. And I'm modifying the code since I didn't cover the case where the owner groups should just be assigned without creating them.

heinrich-ulbricht:
Saw an interesting test by @jensotto - assigning the empty string to an associated group clears this group. Good one, nearly refactored this away.

jansenbe:
hi @heinrich-ulbricht,

Wanted to check if you feel ready to merge the changes you did? I assume they are in your pending PR #2192 ? I did not have time to look into this in detail, but if you feel confident I want to merge this in for the May release. Alternatively we wait for June so more time can be spent on this one

heinrich-ulbricht:
@jansenbe Yes #2192 contains the changes that reflect what was discussed here. My confidence is pretty high and tests are in place to ensure the logic works. So from this standpoint I'd say Go.

Still open are those points in the discussion (but this should be ok for the release):
- discuss how SiteTitle placeholders fit into the picture
- fix the currently failing tests by @jensotto that check certain token handling cases

I still sense a bit of scepticism around the logic we decided to implement so this is the only thing making me think. But we can fine-tune this. And I like to invite everybody implementing their concerns in the form of tests. This is the best way to ensure we covered all requirements.

jansenbe:
Closing this PR as #2192 has been merged as alternative

jensotto:
> Saw an interesting test by @jensotto - assigning the empty string to an associated group clears this group. Good one, nearly refactored this away.
> Edit:
> Setting an associated group to null throws an exception "Value cannot be null." Is this even possible?

@heinrich-ulbricht it's not the operation that sets the associated*group to null that throws an exception, it's the line where the associated*group that just has been set to null is loaded that throws the null reference error:
https://github.com/SharePoint/PnP-Sites-Core/blob/0bdd4ee443dc3f6b662946c5a6be732e5d688cfe/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs#L206-L209

heinrich-ulbricht:
This is "auffallend korrekt". So checking for null before loading should fix this. And commenting in those lines in three instances:

https://github.com/SharePoint/PnP-Sites-Core/blob/0bdd4ee443dc3f6b662946c5a6be732e5d688cfe/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs#L119-L121

PR #2191, opened 2019-04-21
Adding support for provisioning multiple choice field values in DataRow

### LINKED ISSUE #2190: Add support for multi-value choice field values in DataRow
#### Category
- [ ] Bug
- [x] Enhancement

#### Environment
- [ ] SharePoint 2016
- [ ] SharePoint 2013

Provisioning multiple choice field values should be possible.

#### Observed Behavior
Currently the value is set as string, array of strings is not supported for multi-value choice field.

#### Notes
This is necessary to solve https://github.com/SharePoint/PnP-PowerShell/issues/2032

heinrich-ulbricht:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no - yes?
| Related issues?  | fixes #2190, paves way for https://github.com/SharePoint/PnP-PowerShell/issues/2032

Provisioning multiple values in a DataRow for a choice field was not possible. Now it is while preserving compatibility with older templates. New templates can use optional JSON-syntax to specify multiple choice values.

Relates to


heinrich-ulbricht:
Darn. Added wrong check-in. Will create new pull request. Sorry.

heinrich-ulbricht:
Ah sorry. Had to revert the last commit as it went to the wrong branch. It is a separate pull request now.

erwinvanhunen:
Hi Heinrich, we talked this one through with the team, and while we like the added functionality for multichoice fields, we're hovering towards a somewhat different format for the data itself. You assume/require the data to be provided as a JSON string. Of course this works, but we'll end up mixing JSON inside of an element which we _try_ to prevent if possible as this can bring up all kinds of annoyances when it comes to escaping etc. Rather we would like to see an approach which takes the 'standard' / 'old-school' way of doing it in SharePoint using the ;# character combination as a delimiter. 

If you could update your code to use that instead, that would be great! 

We are very much appreciating your work on Core and Powershell! Thank you for all your work!

For now I'll close this PR, awaiting a new PR.

/Erwin

heinrich-ulbricht:
Thanks for your valuable and kind feedback. It has been a long time since I had so much fun diving into an existing code base. Good base but still so much to do :D I also love the insights into your (team's) thinking around certain design decisions, thanks for getting into details.
I'm on it.

heinrich-ulbricht:
@erwinvanhunen Quick question: which format is the correct classic format?

* `value1;#value2`
* `;#value1;#value2;#`

I wanted to go for the first one, but the second one [exists](https://stackoverflow.com/questions/13698050/sharepoint-batch-for-multichoice-field-value-insertion) as well.

And how to handle ;# as part of the value? The JSON approach had this covered but what if the user enters _my;#value_ as value? Should this be encoded as _my;#;#value_ ?

erwinvanhunen:
The delimiter needs to be present between values, but not starting or ending with them. In case people start or end with them you can always trim those away, and then do a simple split on the value I guess?

heinrich-ulbricht:
Sounds good and easy. If this is your impression as well I'd say we go for it.

PR #2192, opened 2019-04-22
Check that exporting and importing of associated groups works

heinrich-ulbricht:
| Q               | A
| --------------- | ---
| New test?      | yes
| Related issues?  | prevent regressions regarding the issue in #2127

Export site as template and apply to newly created site. Check that no more than the expected three groups are created. Also a small fix in other, related code.


jansenbe:
Thanks a ton @heinrich-ulbricht !

jensotto:
@jansenbe @heinrich-ulbricht how can this be merged into dev with at least three failing tests and a fourth test targeting SharePoint 2013 still failing as I pointed out in #2128 and fixed in #2174?
Are there no quality control in what gets added in this project?

heinrich-ulbricht:
@jensotto I had taken a look at the failing tests and my impression was that they check two things: 
- that groups are correctly handled etc. and 
- that the token replacement works while doing this

Due to logic changes discussed in #2174 both aspects are failing and the tests need to be adapted accordingly (which is yet to be done). I decided to ignore these tests for now (but to not remove them to not forget this) until we speak about the token replacement.

With the current logic site title token replacement doesn't make much sense anymore:

- if you want to map to the existing associated groups you don't need to specify the associated group attributes at all - it just works
- when specifying associated group attributes a mapping is done to existing groups with this name; if they don't exist they'll be mapped to the existing associated groups, regardless of their name (this fixes my original problem by the way) - new groups are NOT created in this case
- if you want the groups to be created if not existing you need to list the site groups explicitly

No token needed. Am I totally off here?

About SharePoint 2013 failing: I did not test against SharePoint 2013 which is my fault. I'll have a look at this tomorrow.

jensotto:
@jansenbe @heinrich-ulbricht can we please revert this PR? I don't see how this adds any value as it by my opinion only makes it more difficult to trigger the assigning and possibly creation of associated group(s). It makes the template much hard to use and understand, and authors needs to add the associated groups two places in order for the associated group functionality to trigger. Additionally currently it does not properly work as it compares a de-tokenized group title with a possibly tokenized group title.

There are also a lot of changes to the tests that does not make sense. Why should for example the assertion on line 501 check that the owner group title is different than the web associated owner group title when it clearly needs to be identical?

I have an idea that probably would fix all issues regarding async group rename. I just confirmed that waiting for the async rename to trigger after site creation is not a usable option as it could delay about an hour before it kicks in. 
The solution I propose will use the web property bag vti_createdassociatedgroups, vti_createdassociatedownergroup, vti_createdassociatedmembergroup, and vti_createdassociatedvisitorgroup to get the group id for the default created associated group, and the current associated group.
The associated group tokens ({AssociatedOwnerGroup}, {AssociatedMemberGroup}, and {AssociatedVisitorGroup} would then be used when extracting the template if and only if the associated group id is listed in vti_createdassociatedgroups. When applying the template this token would then never change the existing associated groups. It could be checked for the token directly and not try to parse and check for group title match as the async group rename could have happened in between.
As far as I can see this would solve all the problems that has been reported with this functionality.

The only downside with this is that it will not work with template extraction for SharePoint 2013. vti_createdassociatedgroups only exists for SharePoint 2016 and newer. But I don't think that is too bad as SharePoint 2013 does not have async processing. As a remedy for this it would probably work by using the site title token replacement that I contributed in #2174 

This proposal would require minimal changes in my PR #2174

What do you think?

heinrich-ulbricht:
Interesting. So vti_createdassociatedgroups contains the "original" group IDs, the other properties any non-default groups set by the user?
How does your proposal cover the case where the user created non-default associated groups, exports a template and expects a) the groups to be created and b) them to be assigned as (again non-default) associated groups when applying the template?

jensotto:
It should work just fine by just having the title of the created non-default group(s) specified in the associated groups attribute(s). This would then include the site title token if the group happens to contain the site title. The original associated groups will remain on the site the template is applied to as there is no way to delete groups with the template.
Another option could be to rename the existing associated groups but the template schema does not have that capability, unless we redefine the associated group attributes to mean rename existing associated group if there is one assigned and create if there are none assigned. That is actually a tempting idea.

jensotto:
With a rename group title approach I guess that the async process would change the group name later on if the async group title change has not already happened. It might therefore be difficult to go for that approach.

jensotto:
I've mostly implemented my suggested changes in #2174 now. Just wanted to show my idea so that you can test it.
I need to write a few tests before it is finished, and there is a test that needs to be modified to work with the new extraction method.

jensotto:
I don't understand why I didn't think about this earlier but the solution is even simpler. We could just skip adding the specific associated group attribute if the group is the default one. This way we don't need to specify the associated group token at all.
I'll modify the code tomorrow.

heinrich-ulbricht:
I like it, it seems simple and feels right. And about not thinking about things earlier: this might be associated with late night coding and thinking :D

heinrich-ulbricht:
@jensotto Out of curiosity: how confident are you about using those vti_ groups (like vti_createdassociatedgroups)? Did you use them in the past, discover them while working on this issue or have official sources? They seem to be around since SharePoint 2010. Not many Google hits though.

jensotto:
SharePoint will update those properties every time an associated group is assigned to the web. That is part of the code that runs when setting an associated group. That is also the storage where SharePoint looks to retrieve the associated groups. In other words it's possible to use those properties directly to set the associated groups based on id.

heinrich-ulbricht:
I just tested changing the associated visitor group for a newly created site and this is how the property values look after 2 changes:

```
vti_associateownergroup    3                                                                                                                         
vti_associatemembergroup   5                                                                                                                         
vti_associatevisitorgroup  11                                                                                                                        
vti_createdassociategroups 3;4;5                                                                                                                     
vti_associategroups        5;4;3;10;11                                                                                                               
```

`vti_createdassociategroups` contains the original associated groups and does not change. `vti_associategroups` seem to contain a history of groups that were used in the past. `vti_associate*group` contains the current value. Above you mentioned `vti_createdassociated*group` properties - I don't see those. Or did you mean `vti_associate*group`?

Anyway. Comparing the `vti_associate*group` values with `vti_createdassociategroups` allows to detect if the user changed the groups. This is the base for deciding whether to create an associated attribute with _{Associated*Group}_ placeholder vs. the group name, right?

As a side note: deleting an associated group does not reset the `vti_associate*group` value, it stays as it is even if the group is missing.

jensotto:
Yes, that seems correct. 
I'm sorry for the confusion, I meant `vti_associate*group`. I guess I fell in a copy paste trap.
It's correct thgat those properties will remain even if the group is deleted. Those properties are only updated when the corresponding `vti_associate*group` web property is modified.
The `vti_associategroups` i will get updated with every group you create usiong the UI. It can be modified by clicking the Edit Group Quick Launch in the Settings menu on the groups page.
There are no logic to remove IDs for groups that are deleted.

jensotto:
@heinrich-ulbricht Ok, now I have fixed the code in #2174 so that it does not include the `associated*group` attribute in the template if it is the default created associated group. I also fixed some other small issues and made some improvements around the associated groups token.

jensotto:
sorry. I didn't notice that my PR #2174 did not update with my changes. I guess I have to create a new PR

jensotto:
Here it is #2235

Issue #2194, opened 2019-04-23
Unit tests for XML serialization \ deserialization


s-KaiNet:
#### Category
[x] Enhancement

#### Expected or Desired Behavior
Add more unit tests for upcoming 201903 schema serialization \ deserialization.

jansenbe:
Closing as tests are merged

PR #2225, opened 2019-05-04
Schema 201903 unit tests 

### LINKED ISSUE #2224: Not possible to specify Security for ClientSidePage programmatically
#### Category
[x] Bug

#### Environment
[x] SharePoint 2016

You should be able to set security for ClientSidePage using code object model:
```
var result = new ProvisioningTemplate();
result.ClientSidePages.Add(new Core.Framework.Provisioning.Model.ClientSidePage({...}));
// set security
result.ClientSidePages[0].Security.ClearSubscopes = true;
// etc.
```

#### Observed Behavior
Null reference exception is thrown at line `result.ClientSidePages[0].Security` because `Security` is null. `Security` property has a private `set` modifier and it's not possible to create an instance using the constructor. 

#### Steps to Reproduce
Mentioned in "Expected or Desired Behavior" section

s-KaiNet:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| Related issues?  | fixes #2224 , references #2194 

Added more unit tests for 201913 schema. New nodes introduced in 201913 are covered, 90% of objects and properties from ProvisioningTemplate node are also covered.


jansenbe:
Thanks a lot @s-KaiNet , having more unit tests is super valuable. Great work!

PR #2230, opened 2019-05-08
Using additional headers for HTTP requests

lafe:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes https://github.com/SharePoint/PnP-PowerShell/issues/808, dependency of PR https://github.com/SharePoint/PnP-PowerShell/pull/2068 

In scenarios where an on premises SharePoint server uses Windows Authentication as well as a different authorization provider (e.g. Azure AD), it is not possible to use an Active Directory login to authenticate against SharePoint. As discussed in issue https://github.com/SharePoint/PnP-PowerShell/issues/808, a workaround is to deactivate forms based authentication by passing the `X-FORMS_BASED_AUTH_ACCEPTED` with the value `f` in the header of the request. The first iteration of PR https://github.com/SharePoint/PnP-PowerShell/pull/2068 did that solely within the scope of the PowerShell CmdLets, however this only works with commands that are based on CSOM. If the newer REST API calls are used (in our case, the ALM CmdLets), new HttpClients instances are used that do not inherit the settings (and therefore headers) of the traditional calls. In these instances, the header has to be set again.
This PR implements a generic solution for this issue that uses an optional `Dictionary` that contains additional header that should be added to Web Requests. The related PR https://github.com/SharePoint/PnP-PowerShell/pull/2068 in the PnP-PowerShell project adds the necessary configuration for the `X-FORMS_BASED_AUTH_ACCEPTED` header. 
This generic solution was chosen, because there is currently no approach on how to pass configuration values from the PowerShell command line to fundamental operations in the Sites-Core project. Furthermore, if additional headers might be necessary for other scenarios, these can be added with ease.

### DIFF
--- a/Core/OfficeDevPnP.Core/ALM/AppManager.cs
+++ b/Core/OfficeDevPnP.Core/ALM/AppManager.cs
@@ -12,6 +12,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
+using OfficeDevPnP.Core.Extensions;
 
 namespace OfficeDevPnP.Core.ALM
 {
@@ -21,6 +22,11 @@ namespace OfficeDevPnP.Core.ALM
     /// </summary>
     public class AppManager
     {
+        /// <summary>
+        /// Contains additional headers for the HTTP requests to SharePoint
+        /// </summary>
+        private Dictionary<string, string> AdditionalHeaders { get; }
+
         private ClientContext _context;
         private enum AppManagerAction
         {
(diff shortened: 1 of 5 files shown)


jansenbe:
Hi,

This PR has been open for a long time and we apologize for not having been able to timely process it. Given things have changed a lot over the last year this PR needs to be revisited and updated if the requested change is still relevant. If still relevant then please create a new PR, verified against our dev branch. In the meanwhile we'll be closing this PR, hope you understand.

lafe:
@jansenbe @VesaJuvonen With all due respect: I find this comment really shocking and makes me question contributions to the PnP project. I still have three other open PRs that are at least one year old (the oldest is from 2017). I guess they will all be closed now...
As for this PR: it is still relevant, but the original project has ended and I would have to rework the PR in my free time. Unfortunately, my free time is limited and there are many things that are much more fun to do than to revisit a topic I had already solved over a year ago.

PR #2232, opened 2019-05-09
Fixed NullReferenceException caused by missing cookie

lafe:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | -

This PR fixes an issue that occured when `Credentials` were not set and the cookie was also empty (this is the case, when the `CurrentCredentials` flag is set in the `Connect-PnPOnline` PowerShell CmdLet).
If the `cookieString` is null, it does not make sense to continue executing the method as all subsequent commands expect the `cookieString` to contain a value.


jansenbe:
Thx @lafe !

PR #2235, opened 2019-05-09
Associated groups token

jensotto:
See PR #2174

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
@@ -281,6 +281,40 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
 
                 Assert.IsTrue(template.Security.AdditionalAdministrators.Any());
                 Assert.IsTrue(template.Security.SiteGroups.Any());
+
+                // These three assertions will fail if the site collection does not have the
+                // default groups created during site creation assigned as associated owner group,
+                // associated member group, and associated visitor group.
+                // This is a prerequisite for the site collection used for unit testing purposes.                
+                Assert.AreEqual("{associatedownergroup}", template.Security.AssociatedOwnerGroup, "Associated owner group title is not the Associated Owner Group token.");
+                Assert.AreEqual("{associatedmembergroup}", template.Security.AssociatedMemberGroup, "Associated member group title is not the Associated Owner Group token.");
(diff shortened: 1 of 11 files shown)


jensotto:
@jansenbe

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
```diff
@@ -843,6 +849,44 @@ where user.IsSiteAdmin
...
+                    throw new ArgumentOutOfRangeException(nameof(groupType));
+            }
+
+            string createdAssociateGroupsKey = "vti_createdassociategroups";
+            string associatedGroupKey = $"vti_associate{groupTypeText}group";
+
+            if (web.AllProperties.FieldValues.ContainsKey(createdAssociateGroupsKey) &&
+                web.AllProperties.FieldValues.ContainsKey(associatedGroupKey))
+            {
+                string associatedOwnerGroupId = (string)web.AllProperties.FieldValues[associatedGroupKey];
```
heinrich-ulbricht: `associatedOwnerGroupId` should be more generic, like `associatedGroupId`

Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectSiteSecurityTests.cs
```diff
@@ -281,6 +281,40 @@ public void CanCreateEntities2()
...
+                creationInfo.IncludeSiteGroups = true;
+                creationInfo.IncludeDefaultAssociatedGroups = true;
+                var template = new ProvisioningTemplate();
+                template = new ObjectSiteSecurity().ExtractObjects(web, template, creationInfo);
+
+                ctx.Load(web,
+                    w => w.AssociatedOwnerGroup.Title,
+                    w => w.AssociatedMemberGroup.Title,
+                    w => w.AssociatedVisitorGroup.Title);
+                ctx.ExecuteQuery();
```
heinrich-ulbricht: ExecuteQueryRetry?

heinrich-ulbricht:
It got rather quiet since our burst of activity. Since the conference time is over I think we can ramp up activity again :) This pull request is rather important I think. @jansenbe Do you see any chance of reviewing and merging this (replacing my implementation)?

heinrich-ulbricht:
@jansenbe This longer we wait the more complicated it gets. What do you suggest? The changes by @jensotto seemed pretty good.

jansenbe:
We're working on API updates that allow us to detect when the "async" part of the provisioning flow is done...with that in place the current code should be more reliable. Let's re-evaluate this problem once we're at that stage. So for now I'll close this PR, we can for sure re-open it afterward if needed. Thanks for your understanding.

PR #2241, opened 2019-05-14
Fix: Additional null checks while provisioning team

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes NA

1) This PR adds additional null check while provisioning teams with additional settings like FunSettings, GuestSettings etc. 

2) Additionally, it removes compiler directives which are not needed as we already handle it "above".

3) If the Giphy content rating is set to Strict, that value is not respected, so fixed that issue with static class.

4) Added support for users with special characters in email address.

This PR fixes null issue when you try to provision a team from PnP template which is as below:

```
<pnp:Teams>    
    <pnp:Team DisplayName="Sample Team 01" Description="This is just a sample Team 01" Visibility="Public" Specialization="None">
      <pnp:Security>
        <pnp:Owners ClearExistingItems="true">
          <pnp:User UserPrincipalName="admin@tenant.onmicrosoft.com" />
          <pnp:User UserPrincipalName="meganb@tenant.onmicrosoft.com" />
        </pnp:Owners>
        <pnp:Members ClearExistingItems="true">
          <pnp:User UserPrincipalName="admin@tenant.onmicrosoft.com" />
          <pnp:User UserPrincipalName="meganb@tenant.onmicrosoft.com" />
        </pnp:Members>
      </pnp:Security>
    </pnp:Team>
</pnp:Teams>	
```

When i try to provision a Team using above template, the Team is not getting created, error is not thrown either but debugging the code, it gives a null reference error. This PR fixes it.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
@@ -1,23 +1,19 @@
 ﻿#if !ONPREMISES
-using Microsoft.SharePoint.Client;
+using Microsoft.Online.SharePoint.TenantAdministration;
+using Newtonsoft.Json.Linq;
 using OfficeDevPnP.Core.Diagnostics;
+using OfficeDevPnP.Core.Framework.Provisioning.Connectors;
 using OfficeDevPnP.Core.Framework.Provisioning.Model;
+using OfficeDevPnP.Core.Framework.Provisioning.Model.Teams;
+using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.Utilities;
+using OfficeDevPnP.Core.Utilities;
+using OfficeDevPnP.Core.Utilities.Graph;
 using System;
+using System.IO;
 using System.Linq;
 using System.Net.Http.Headers;
-using Newtonsoft.Json.Linq;
-using OfficeDevPnP.Core.Framework.Provisioning.Model.Teams;
-using OfficeDevPnP.Core.Utilities;
-using System.Net;
-using Newtonsoft.Json;
-using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using System.Web;
-using System.Net.Http;
-using Microsoft.Online.SharePoint.TenantAdministration;
-using OfficeDevPnP.Core.Utilities.Graph;
-using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.Utilities;
-using OfficeDevPnP.Core.Framework.Provisioning.Connectors;
-using System.IO;
 
 namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 {
@@ -55,8 +51,8 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                 // Check if the Group exists
                 if (GroupExistsById(scope, parsedGroupId, accessToken))
                 {
-                    // Then promote the Group into a Team or update it, if it already exists
-                    teamId = CreateOrUpdateTeamFromGroup(scope, team, parser, parsedGroupId, accessToken);
+                    // Then promote the Group into a Team or update it, if it already exists. Patching a team doesn't return an ID, so use the parsedGroupId directly (teamId and groupId are the same). 
+                    teamId = CreateOrUpdateTeamFromGroup(scope, team, parser, parsedGroupId, accessToken) ?? parsedGroupId;
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


jackpoz:
I also receive a null reference exception at https://github.com/SharePoint/PnP-Sites-Core/blob/master/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs#L870 because PnPProvisioningContext.Current is null:
image

Does this PR handle this case too ? (I'm not even sure why it's null or how the rest of the provisioning engine handles the case of a null PnPProvisioningContext.Current)

gautamdsheth:
@jackpoz  - good catch, it didn't handle that ! I have added that check now via a new commit. Thanks.

jackpoz:
do you know how PnPProvisioningContext.Current is provided ? Is it the caller that has create a new PnPProvisioningContext or ?
Are you using Apply-PnPTenantTemplate or your own .NET app ? Could you try with Apply-PnPTenantTemplate too ?

jackpoz:
After applying https://github.com/SharePoint/PnP-PowerShell/pull/2099 and this PR, Apply-PnPTenantTemplate can create a Team without throwing any exception 👍

PaoloPia:
Hi,
Thanks for your contribution. About the PnPProvisioningContext object, here (https://www.youtube.com/watch?v=n6Mh9AzO9A8&t=3s) you can find a short video where I explain the purpose and how to use it. It is supposed to be provided by the caller in .NET code, and it is handled by PnP PowerShell automatically. About the suggested changes, I'm evaluating them. Thanks.

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
```diff
@@ -594,7 +594,7 @@ private static bool SetGroupSecurity(PnPMonitoredScope scope, Team team, string
         /// <returns>Whether the Channels have been provisioned or not</returns>
         private static bool SetTeamChannels(PnPMonitoredScope scope, TokenParser parser, Team team, string teamId, string accessToken)
         {
-            if (team.Channels != null)
+            if (team.Channels.Count > 0)
```
PaoloPia: Hi, what is the purpose of this test? If the Channels collection is not null but still empty the foreach clause will be skipped. On the contrary if Channels is null and we test the Count property, we will get a null reference exception.

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
```diff
@@ -622,66 +622,68 @@ private static bool SetTeamChannels(PnPMonitoredScope scope, TokenParser parser,
                     if (channelId == null) return false;
 
                     // If there are any Tabs for the current channel
-                    if (channel.Tabs == null || !channel.Tabs.Any()) continue;
-
-                    foreach (var tab in channel.Tabs)
+                    if (channel.Tabs.Count > 0)
```
PaoloPia: Same here, the foreach over an empty collection will simply skip it. I think we don't need to test Count > 0.

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
```diff
@@ -622,66 +622,68 @@ private static bool SetTeamChannels(PnPMonitoredScope scope, TokenParser parser,
...
                     }
 
                     // TODO: Handle TabResources
                     // We need to define a "schema" for their settings
 
                     // If there are any messages for the current channel
-                    if (channel.Messages == null || !channel.Messages.Any()) continue;
-
-                    foreach (var message in channel.Messages)
+                    if (channel.Messages.Count > 0)
```
PaoloPia: And the same note applies here.

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
```diff
@@ -622,66 +622,68 @@ private static bool SetTeamChannels(PnPMonitoredScope scope, TokenParser parser,
...
-                            accessToken,
-                            "NameAlreadyExists",
-                            CoreResources.Provisioning_ObjectHandlers_Teams_Team_TabAlreadyExists,
-                            "displayName",
-                            tab.DisplayName,
-                            CoreResources.Provisioning_ObjectHandlers_Teams_Team_ProvisioningError,
-                            canPatch: false);
-
-                        if (tabId == null) return false;
+                                tab.DisplayName,
```
PaoloPia: Could you please explain the purpose of this change?

jackpoz:
> and it is handled by PnP PowerShell automatically

Erwin fixed that part in https://github.com/SharePoint/PnP-PowerShell/commit/e3b3a4e2470190af599229e33c794033cdc5f813 so that kind of explained why it didn't work before

gautamdsheth:
closing this PR , had some merge conflicts :(

PR #2049, opened 2019-01-25
Added UpdateChildren property to FieldRef to update/not-update inheriting content types

maxali:
…contenttypes, fieldref

| Q               | A
| --------------- | ---
| Bug fix?        | yes?
| New feature?    | yes?
| New sample?      | no
| Related issues?  | fixes #2045 

#### What's in this Pull Request?

Added `UpdateChildren` property to FieldRef. This allows to update/not update inheriting content types as in the [documentation](https://docs.microsoft.com/en-us/previous-versions/office/sharepoint-server/ee544078(v%3Doffice.15)).

Currently if you update FieldRef in `Document` or `Item` content types as of version 3.5.1901, you get an error like this: 
image

This is because [contentType.update(true)](https://github.com/SharePoint/PnP-Sites-Core/blob/14aaea787a160edbfa89e6ab132340da6eb7a5cf/Core/OfficeDevPnP.Core/Extensions/FieldAndContentTypeExtensions.cs#L1038) is hard-coded to pass `true` to parameter `updateChildren`. This will always force SharePoint to update inheriting content types which can be read-only or sealed. 

This PR gives the flexibility to send `UpdateChildren="False"` parameter for such content types.

Example:
```xml
        <pnp:ContentType ID="0x0101" Name="Document" Description="Create new document">
          <pnp:FieldRefs>
            <pnp:FieldRef ID="{61e031fe-3c5e-4bbb-b451-868721fbbc5a}" UpdateChildren="false" Required="false" Name="CustomField1" />
          </pnp:FieldRefs>
        </pnp:ContentType>
```

PR #2270, opened 2019-05-27
Content type update children

### LINKED ISSUE #2251: UpdateChildren on FieldRef doesn't appear to work.  It still errors trying to update child content types.
#### Category
[X] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ ] SharePoint 2013

As per #2049 it seems this should work.

Setting `UpdateChildren="false"` should not try an update child content types

#### Observed Behavior
An error is returned notifying me that child content types are sealed, indicating that it's trying to update all children.

#### Steps to Reproduce

**Template:**
```
 <pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2019/03/ProvisioningSchema">
       <pnp:Preferences Generator="OfficeDevPnP.Core, Version=3.9.1905.2, Culture=neutral,PublicKeyToken=5e633289e95c321a" />
      <pnp:Templates>
```
**Content type section:**

```

    <pnp:ContentType ID="0x0101" Name="Document">
      <pnp:FieldRefs>
        <pnp:FieldRef ID="{48D055D6-6502-4B4E-9C9F-9F4A1AF18143}" Name="Owner" UpdateChildren="false" />
      </pnp:FieldRefs>
    </pnp:ContentType>
```

**Version:**
image

**Result**
```
Apply-PnPProvisioningTemplate : The content type "Office Data Connection File" at "/sites/test1" is read only.
The content type "Universal Data Connection File" at "/sites/test1" is read only.
The content type "Design File" at "/sites/test1" is sealed.
The content type "Unknown Document Type" at "/sites/test1" is read only.
Access is denied. (Exception from HRESULT: 0x80070005 (E_ACCESSDENIED))
Access is denied. (Exception from HRESULT: 0x80070005 (E_ACCESSDENIED))
At line:5 char:1
The error is being thrown the same as it did in previous versions, because it's trying to update child content types of Document which are sealed.
```
Something I'm missing?

SteveClements:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #2251 #2049

UpdateChildren was added to FieldRefs in Content Types.  According to [this](https://github.com/SharePoint/PnP-Sites-Core/pull/2049#issuecomment-494748756) comment the original PR wasn't merged.

This includes the changes in the original PR, plus an additional `AddFieldById` overload.

This also wasn't the whole story as any other changes to the content type would mean `Update` is called, plus if FieldRefs are in a different order again `Update` is called with updateChildren true.
I've defaulted the first `Update`, which is CT properties only to `false`
I've added a slightly questionable workaround where I check that is there are changes to the FieldRefs it will only update children if one has `UpdateChildren` set to true

However...this still isn't 100%, as `UpdateChildren` isn't required in the XML, the default value on the `FieldRef` property is false, but the default value of `UpdateChildren` in the engine wants to be true.
I did try setting it to be true by default, but that didn't work.
`public bool UpdateChildren { get; set; } = true;`

Ideally, `UpdateChildren` should be a property at the Content Type level rather than the field ref I think or both the CT and the FieldRef (but I can't think of a scenario where I'd only want certain fields to be pushed down, plus with the re-order causing and update, if pushed and the fields weren't present in the child, I'd expect and error).  So perhaps at the CT level is best and if you only want certain fields pushed, you can have two CT elements in your template doing different things.

At least for now at least we have a working solution, with a caveat!  To force and update of child content types you must set at least one field ref's `UpdateChildren` property to `true`

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/FieldAndContentTypeExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/FieldAndContentTypeExtensions.cs
@@ -968,60 +968,64 @@ namespace Microsoft.SharePoint.Client
             return true;
         }
 
-        /// <summary>
-        /// Associates field to content type
-        /// </summary>
-        /// <param name="contentType">Content Type to add the field to</param>
-        /// <param name="fieldId">String representation of the id of the field (=Guid)</param>
-        /// <param name="required">True if the field is required</param>
-        /// <param name="hidden">True if the field is hidden</param>
-        public static void AddFieldById(this ContentType contentType, string fieldId, bool required = false, bool hidden = false)
+		/// <summary>
+		/// Associates field to content type
+		/// </summary>
+		/// <param name="contentType">Content Type to add the field to</param>
+		/// <param name="fieldId">String representation of the id of the field (=Guid)</param>
+		/// <param name="required">True if the field is required</param>
+		/// <param name="hidden">True if the field is hidden</param>
+		/// <param name="updateChildren">True to update content types that inherit from the content type; otherwise, false.</param>
+		public static void AddFieldById(this ContentType contentType, string fieldId, bool required = false, bool hidden = false, bool updateChildren = true)
         {
-            AddFieldById(contentType, Guid.Parse(fieldId), required, hidden);
+            AddFieldById(contentType, Guid.Parse(fieldId), required, hidden, updateChildren);
         }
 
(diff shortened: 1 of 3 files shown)


PaoloPia:
Hi @SteveClements,
Thanks for your contribution. This PR has just been merged!

PR #2298, opened 2019-06-14
Feature: Added support to delete modern sites

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | fixes NA

This PR adds support to delete modern communication or modern team site(groupless as well as backed by O365 group) 
It is based on [create modern SharePoint site](https://docs.microsoft.com/en-us/sharepoint/dev/apis/site-creation-rest) guidance.

### DIFF
--- a/Core/OfficeDevPnP.Core/Sites/SiteCollection.cs
+++ b/Core/OfficeDevPnP.Core/Sites/SiteCollection.cs
@@ -627,7 +627,7 @@ namespace OfficeDevPnP.Core.Sites
         /// Enable Microsoft Teams team in an O365 group connected team site
         /// Will also enable it on a newly Groupified classic site
         /// </summary>
-        /// <param name="context"></param>
+        /// <param name="context">Context to operate against</param>
         /// <returns></returns>
         public static async Task<string> TeamifySiteAsync(ClientContext context)
         {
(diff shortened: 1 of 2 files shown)


Core/OfficeDevPnP.Core/Sites/SiteCollection.cs
```diff
@@ -654,6 +654,45 @@ public static async Task<string> TeamifySiteAsync(ClientContext context)
...
+
+                var parsedResult = JObject.Parse(result);
+
+                siteDeleted = Convert.ToBoolean(parsedResult["odata.null"]);
+
+                return await Task.Run(() => siteDeleted);
+            }
+            else
+            {
+                throw new Exception("Only deletion of Communication site or Modern team site is supported by this method.");
```
mdissel: Why not automatically delete "old" site?
Maybe make a DeleteModernSite that only deletes a Modern site and a general DeleteSite that always deletes the site regardless of the type of site (and linked group).

gautamdsheth: You can already delete the sites of all templates using TenantExtensions. This PR adds an additional method in the Site collection context instead of Tenant context. Also, we are using the newer APIs which were specifically made for Modern sites.

Issue #1601, opened 2018-02-21
List View WebPart does not keep selected View when provisioned


erwinvanhunen:
_From @sivamsolai on February 16, 2018 10:8_

Hi Team 

The list view part not maintaining selected view on Home page when provisioned. We added the calendar list view web part and “calendar” view is selected. But when provisioned not maintaining view.  
**Schema**
`<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2018/01/ProvisioningSchema">` 

**Web part**
` <pnp:WebPart Title="Calendar" Row="1" Column="1">
              <pnp:Contents>
                <WebPart xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://schemas.microsoft.com/WebPart/v2">
                  <Title>Calendar</Title>
                  <FrameType>Default</FrameType>
                  <Description />
                  <IsIncluded>true</IsIncluded>
                  <ZoneID>wpz</ZoneID>
                  <PartOrder>0</PartOrder>
                  <FrameState>Normal</FrameState>
                  <Height />
                  <Width />
                  <AllowRemove>true</AllowRemove>
                  <AllowZoneChange>true</AllowZoneChange>
                  <AllowMinimize>true</AllowMinimize>
                  <AllowConnect>true</AllowConnect>
                  <AllowEdit>true</AllowEdit>
                  <AllowHide>true</AllowHide>
                  <IsVisible>true</IsVisible>
                  <DetailLink>{site}/Lists/Calendar</DetailLink>
                  <HelpLink />
                  <HelpMode>Modeless</HelpMode>
                  <Dir>Default</Dir>
                  <PartImageSmall />
                  <MissingAssembly>Cannot import this Web Part.</MissingAssembly>
                  <PartImageLarge>/_layouts/15/images/itevent.png?rev=40</PartImageLarge>
                  <IsIncludedFilter />
                  <Assembly>Microsoft.SharePoint, Version=16.0.0.0, Culture=neutral, PublicKeyToken=71e9bce111e9429c</Assembly>
                  <TypeName>Microsoft.SharePoint.WebPartPages.ListViewWebPart</TypeName>
                  <WebId xmlns="http://schemas.microsoft.com/WebPart/v2/ListView">00000000-0000-0000-0000-000000000000</WebId>
[...]

PR #2304, opened 2019-06-20
Added BaseViewID into ViewCreationInformation

StaffanNelemans:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no 
| New sample?      | no 
| Related issues?  | #1601

All views created are having BaseViewId = 1, even though BaseViewID is in the extracted template. For example an standard Calendar List.


jansenbe:
Thanks a lot @StaffanNelemans , nice fix. I did some small tweaks to make this change only be included for SP2019 and SPO as the needed clientside OM does not have this option for SP2013 and SP2016

PR #2308, opened 2019-06-25
CAML improvements

kirschem:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | -

This PR contains a little improvement for the CAML class. When doing a CAML query that contains a In-comparision, one would need to forge most of the query by himself with the current class:
```csharp
var fieldRef = CAML.FieldRef("ID")+"<Values><Value Type="Number">1</Value><Value Type="Number">2</Value></Values>";
var caml = CAML.In(fieldRef);
```
There is no method for creating a Value node without a FieldRef node. Furthermore the Values node also needs to be created manually.

I've added an overload for the In method, accepting a FieldRef string and an array for Value strings. In Addition, I've added a method "Value" to generate the string for a Value node.

Usage:
```csharp
var fieldRef = CAML.FieldRef("ID");
var values = new string[] { CAML.Value("1", "Integer"), CAML.Value("2", "Integer") };
var caml = CAML.In(fieldRef, values);
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,3 +178,6 @@ FakesAssemblies/
 
 # Visual Studio 2015 cache/options directory
 .vs/
+
+# VS Code Settings
+.vscode/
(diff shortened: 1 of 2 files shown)


jansenbe:
Thx @kirschem for sharing this.

PR #2318, opened 2019-07-04
Allow ZoneEmphasis in ClientSidePage to be undefined

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes https://github.com/SharePoint/PnP-PowerShell/issues/2141

ZoneEmphasis in some cases returns undefined when requesting a ClientSidePage. This code change will ensure the PnP Core library can deal with this whereas now it would crash. I ran into this issue while fixing issue https://github.com/SharePoint/PnP-PowerShell/issues/2141 where requesting a clientside page on a subsite would return the client page but with the zoneemphasis information set to "undefined" which caused the code to crash trying to JSONConvert that into an int.


jansenbe:
thx @KoenZomers !

PR #2321, opened 2019-07-08
Added Built in CT Html Page layout

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Consider the following scenario.

1. Activate Publishing Infrastructure
2. Deactivate Publishing Infrastructure
3. Export site collection
4. Import template to site collection where Publishing is NOT activated

The import will fail because the HtmlPageLayout content type cannot be created because the parent content type doesn't exist.
This PR excludes the HtmlPageLayout from being exported by adding it to built in content types.


jansenbe:
Thx @SchauDK

PR #2340, opened 2019-07-26
Minor bugfixes for Channels and Tabs

PBLPointwork:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | NA

This PR includes minor bugfixes CreateTeam From ProvisioningSchema 

- Checking if team.Security section is configured, if it isn't set this part is skipped.
- Included the TokenParser as parameter to Tabs functionality, allowing to use parameters for Tabs
- In the UpdateTeamChannel made a check for name 'General' as it is not allowed to make PATCH to the General Channel

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectTeams.cs
@@ -698,6 +698,9 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 
             foreach (var tab in tabs)
             {
+                // Avoid ActivityLimitReached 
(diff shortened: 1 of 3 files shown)


jansenbe:
Thx for your contribution @PBLPointwork

LauraKokkarinen:
Hi Peter!

Thanks a lot for these fixes and improvements! Especially checking if the channel is the default General channel was a very good one. That made me think about it a bit more:

The channel isn't actually always called "General"; it can also be something else depending on localization. For example in Finnish, the default channel is called "Yleinen". I wonder if there is a more reliable way to do that so we wouldn't need to worry about the selected language... will definitely need to think about this.

Anyway, still a major improvement to what we had before so big thanks! Keep up the good work. :)

Laura

PR #2363, opened 2019-08-19
Feature: S2S user delegation in Hightrust app scenario

tmeckel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |  yes
| New sample?      | no
| Related issues?  |  n/a

This PR contains an enhancement whith which it's possible to create an App Delegation token in an Hightrust configuration to act as specific user (behalf-of) within a SharePoint connection. Up to now this scenario was only possible with an `WindowsIdentity` object instance available, which is not the case if the addin is deployed in an environment which isn't connected to an Active Directory environment (e.g. Azue WebApp) and is employing other authentication themes like JWT Bearer or SAML tokens. 

This PR now provides new methods on `Core/OfficeDevPnP.Core/AuthenticationManager.cs` which allows to pass the SID or the UPN for an user for creating an impersonation (delegation) token. The SID is just as unavailable in the above scenario as the `WindowsIdentity` object instance but the UPN typically is, because the UPN is normally used in JWT or SAML token to identify an incoming user. 

With this extension it is possible to host a SharePoint Add in a "domain-less" environment, like as an Azure WebApp, utilize modern authentication methods and still perform impersonation on a SharePoint system in an Active Directory domain.

### DIFF
--- a/Core/OfficeDevPnP.Core/AuthenticationManager.cs
+++ b/Core/OfficeDevPnP.Core/AuthenticationManager.cs
@@ -613,7 +613,7 @@ namespace OfficeDevPnP.Core
             // Configure the handler to generate the Bearer Access Token on each request and add it to the request
             clientContext.ExecutingWebRequest += (sender, args) =>
             {
-                var accessToken = TokenHelper.GetS2SAccessTokenWithUserName(siteUri, loginName);
+                var accessToken = TokenHelper.GetS2SAccessTokenWithWindowsUserName(siteUri, loginName);
                 args.WebRequestExecutor.RequestHeaders["Authorization"] = "Bearer " + accessToken;
             };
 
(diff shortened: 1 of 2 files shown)


Core/OfficeDevPnP.Core/AuthenticationManager.cs
```diff
@@ -520,6 +520,105 @@ public ClientContext GetHighTrustCertificateAppOnlyAuthenticatedContext(string s
...
+        /// <param name="loginName">
+        /// Name of the user (login name) on whose behalf to create the access token.
+        /// Supported input formats are SID and User Principal Name (UPN).
+        /// If the parameter is left empty (including null) an App Only Context will be created.
+        /// </param>
+        /// <returns>Authenticated SharePoint ClientContext</returns>
+        public ClientContext GetHighTrustCertificateAppAuthenticatedContext(string siteUrl, string clientId, string certificatePath, string certificatePassword, string certificateIssuerId, string loginName)
+        {
+            var certPassword = Utilities.EncryptionUtility.ToSecureString(certificatePassword);
+            return GetHighTrustCertificateAppAuthenticatedContext(siteUrl, clientId, certificatePath, certPassword, certificateIssuerId, loginName);
```
gszdev: @tmeckel 
There's a little compile error.
You are using a not defined variable named "certPassword" (Line 541 ) but the parameter name of the method is "certificatePassword" (line 538).

jansenbe:
Thanks @tmeckel. Just merged this one, did revert the public method rename you did to avoid breaking folks that where already using this.

tmeckel:
> 
> 
> Thanks @tmeckel. Just merged this one, did revert the public method rename you did to avoid breaking folks that were already using this.

@jansenbe Thanks for merging my PR and spotting the breaking change and reverting it!

PR #2366, opened 2019-08-20
Fix list view localization (when destination single Language and PnP-Template Multilanguage)

### LINKED ISSUE #2365: ListViewLocalization sets wrong Language when destination Site has only one Language

#### Category
[x] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ ] SharePoint 2013

We have a PnP-Template with Ressources including multiple Languages but destination sites has only one language active we would like to have the correct translation on the View.

#### Observed Behavior
We will have the Last Language in Resx-Files apply on View which might not be the correct Language.

#### Steps to Reproduce
Apply a Template with Resx-Ressources to a Site having one Language without 
Handlers.SupportedUILanguages and your List Views will have the wrong Language.

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |   yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #2365 

The fix checks if destination site IsMultilingual and skips localization if not. 
If the Site IsMultilingual=true, it checks if the language from ressourcefile is in SupportedUILanguageIds of the web and applys only if they exist.

### DIFF
--- a/Core/OfficeDevPnP.Core/CoreResources.Designer.cs
+++ b/Core/OfficeDevPnP.Core/CoreResources.Designer.cs
@@ -1020,6 +1020,24 @@ namespace OfficeDevPnP.Core {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Translation for default language of site is missing.
+        /// </summary>
+        internal static string Provisioning_Extensions_ViewLocalization_DefaultLngMissing_Skip {
+            get {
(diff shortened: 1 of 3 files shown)


jansenbe:
Thanks @czullu . Good change, I've replicated this model as well for the web part and navigation translations.

PR #2368, opened 2019-08-26
Don't pass formatted string to logger

SchauDK:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

In the ExtensibilityManager class a formatted string is passed to the Log.Error function.
The Log.Error function is doing string.Format(), so we should leave it the Error function to format the string to avoind getting "Input string was not in a correct format." (in case the string contains {} characters)


jansenbe:
Thx @SchauDK

PR #2377, opened 2019-09-03
Added GetNestedUnifiedGroupMembers method

anoopt:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no 
| Related issues?  | NA

Add a method called `GetNestedUnifiedGroupMembers` which will get all the users in a group including the users that are present in nested groups.

Instead of editing and disturbing the existing `GetUnifiedGroupMembers` and `GenerateGraphUserCollection` I have added separate methods. However, if this functionality can be implemented in the same methods please feel free to or let me.

#### New methods added are:
- GetNestedUnifiedGroupMembers (line 948)
- GenerateNestedGraphUserCollection (line 1108)
- GenerateGraphGroupCollection (line 1141)

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Graph/UnifiedGroupsUtility.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Graph/UnifiedGroupsUtility.cs
@@ -937,6 +937,79 @@ namespace OfficeDevPnP.Core.Framework.Graph
         }
 
         /// <summary>
+        /// Returns all the Members of an Office 365 group (including nested groups).
+        /// </summary>
+        /// <param name="group"></param>
+        /// <param name="accessToken"></param>
+        /// <param name="retryCount"></param>
+        /// <param name="delay"></param>
+        /// <returns></returns>
+
+        public static List<UnifiedGroupUser> GetNestedUnifiedGroupMembers(UnifiedGroupEntity group, string accessToken, int retryCount = 10, int delay = 500)
+        {
+            List<UnifiedGroupUser> unifiedGroupUsers = new List<UnifiedGroupUser>();
+            List<User> unifiedGroupGraphUsers = null;
+            IGroupMembersCollectionWithReferencesPage groupUsers = null;
+
+            if (String.IsNullOrEmpty(accessToken))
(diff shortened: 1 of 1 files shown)


LauraKokkarinen:
Hi @anoopt ,

While reviewing your PR, I noticed a couple of things I'd like to get a bit more information about:

- Office 365 groups a.k.a. Unified groups can't be nested with one another. In light of this information, what is the purpose of this code?

- The code isn't actually used by the engine (GetNestedUnifiedGroupMembers has 0 references). Would you mind telling me what you are planning to use it for?

Thanks!

Laura

anoopt:
Hi @LauraKokkarinen ,

Recently we had a requirement wherein we had to read all the users from an an Azure AD group including the users in a nested group. Once all the users were retrieved we had to use them in Flow and start an approval process. 
I used the `GetUnifiedGroupMembers` method for this and realized that I was not able to get the users from nested groups. Hence I created the `GetNestedUnifiedGroupMembers` method. 
I thought this might be useful for anyone who would need all the users in an Azure AD group including nested users and hence I submitted a PR.
Hope this answers the questions :)

Thanks,
Anoop

LauraKokkarinen:
Hi @anoopt ,

Sounds good! I think these methods can absolutely be useful to other people as well, so definitely looking forward to merging them. However, I'd like to request a couple of changes to your PR before I merge it:

1. Because you are actually not checking through unified groups here (because unified groups can't be nested) but regular Azure AD security groups, could you rename the methods and move them to a different file? That'd give a clearer indication of what the code is about and avoid confusion other people might have between unified and regular security groups.
2. Could you write the following kind of a unit test for these methods, so it will be easy to make sure the code works (now initially but also if someone later wants to make changes to it)?
    - Initialize: Create users and nested groups in Azure AD for testing purposes, and add those users to the groups
    - TestMethod: Make a call to your code to retrieve the users from groups. Do an assert to check if you got the expected result -> the test passed/failed.
    - TestCleanup: Remove the Azure AD groups and users you created during initialization.

Thanks a lot! :)

Laura

Issue #2254, opened 2019-05-21
Web Part background white in Page Section with Zone Emphasis


bogeorge:
Thank you for reporting an issue or suggesting an enhancement. We appreciate your feedback - to help the team to understand your needs, please complete the below template to ensure we have the necessary details to assist you. If you have a actual question, we would ask you to use [SharePoint Developer Group](http://aka.ms/sppnp-community) at Microsoft Tech Community. Thanks!

#### Category
[ X ] Bug
[ ] Enhancement

#### Environment
[ X ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

If SharePoint on-premises, what's exact CU version: 

#### Expected or Desired Behavior
When applying a provisioning template to a ClientSidePage that is using the Background Emphasis for a Section the web parts should respect the background color.

#### Observed Behavior
When applying a provisioning template to a ClientSidePage that is using the Background Emphasis for a Section the web parts (NewsReel specifically) actually get a white background that overrides the section background color.

This appears to come from classes assigned to a DIV that has none in the original HTML.

#### Steps to Reproduce
- Create a simple modern page with the section background set to neutral and add a News web part.
This screen shows News with a grey background (neutral) matching the section.  The HTML on the right is highlighting a div that is different after the template is applied (2nd screen shot).
![SourceHome](https://user-images.githubusercontent.com/10885472/58121814-3ee62a00-7bd6-11e9-9ee5-51f2157b594a.png)

- Un-comment and use ClientSidePagesTests.cs - PageTest test method to extract a template for the site page.  You can also use PnP PowerShell or other methods

- Un-comment and use ClientSidePageTests.cs - Page2Test method to apply the template to a site (ideally a different site so you can compare).

This screen shows News with a white background (neutral) matching the section.  The HTML on the right is highlighting a div that now has a class root-76 that is making the background white for the webpart overridding the section background
![DestinationHome](https://user-images.githubusercontent.com/10885472/58121821-4279b100-7bd6-11e9-9954-0f114f160fe9.png)

Additional Notes:
- I have attempted to debug and break in ClientSidePage.Save just before item.Update.  I essentially tried to save the original value pulled from CanvasContent1 and then set it back unaltered by code.  Doing this still resulted in a white background which gives me the impression that server side code may be responsible for the issue.

![code](https://user-images.githubusercontent.com/10885472/58122260-32160600-7bd7-11e9-890d-608cf2001778.png)

[...]

PR #2389, opened 2019-09-04
Fix for issues between two versions of News (older NewsreelWebPart vs newer NewsWebPart)

### LINKED ISSUE #2320: Provisioned News web part is not filtering correctly base on the configured page property filter
Thank you for reporting an issue or suggesting an enhancement. We appreciate your feedback - to help the team to understand your needs, please complete the below template to ensure we have the necessary details to assist you. If you have a actual question, we would ask you to use [SharePoint Developer Group](http://aka.ms/sppnp-community) at Microsoft Tech Community. Thanks!

*Please navigate to the appropriate repository by clicking on the link below and create your issue there. You can delete this section after you have navigated to the correct repository.*
- PnP Samples - https://github.com/OfficeDev/PnP
- PnPjs - https://github.com/pnp/pnpjs
- PnP PowerShell - https://github.com/OfficeDev/PnP-PowerShell

#### Category
[ X ] Bug
[ ] Enhancement

#### Environment
[ ] SharePoint 2016
[ ] SharePoint 2013

Provisioning a ClientSidePage configured with a News Web part that contains configured Page Property filter should filter based on that filter.

#### Observed Behavior

- If the News Web part is the first News web part provisioned on the page it does not filter and requires it to be manually removed and then re-added.

- If the News Web part is NOT the first News web part provisioned on the page it does appear to filter correctly but only until the first web part is removed.  Then it filters incorrectly (or not at all)

Both of these "Featured Brand News" web parts have the EXACT same JsonControlData except instanceId.  When provisioned they show different results, when using Export-PnPClientSidePage the output the exact same output.
image

#### Steps to Reproduce

- On any Site Pages library add additional column to use for filtering.  In my case I'm using a Managed Metadata Column.
- Add 1 or more News Articles and tag them with unique terms like Featured or some value you are going to filter on.  Also Add some that are not tagged, make sure these are "newer" than those you tagged.
- On the home page (or any page), Add a News web part and delete everything else (just to keep it clean).
- Configure the News web part to filter on your tag such as filtered.  This **should** be successful in filtering your News Articles to just those you tagged.
- Export the page using either Export-ClientSidePage or Get-PnPProvisioningTemplate, we just want the Xml for the pnp:ClientSidePage.  I would maybe give it another PageName so we don't overwrite the source page it was based on.
- Use Apply-PnPProvisioningTemplate to apply the template to the site it was generated from.
- This **should** "break" in that the filter doesn't work as it did in the source page and seems to just show the most recent news.
- An additional test I tried was to copy the pnp:CanvasControl for the NewsReal and duplicate it and then change the instanceId slightly.  This is what created the screenshot above where instance 1 is "broken" but instance 2 is working.  I decided to this after learning that Add-PnPClientSideWebPart had the same result.  If that is executed and this is the first and only News web part it will be broken, however if you execute it again, the second copy worked.

#### Other Notes

I have been using TestMethods in ClientSidePagesTests.cs and trying various updates to ObjectClientSidePages.cs to see if I can figure out a way to fix this but so far have come up short.  The JsonControlData that is being set on myWebPart.PropertiesJson looks correct (based on what was exported).

Please let me know if there is any additional information you need from me.  I'll keep digging on my end.

bogeorge:
**|** Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #2320 and #2254 

This is a fix for a new version of the "NewsReel" (Guid: a5df8fdf-b508-4b66-98a6-d83bc2597f63) that is now called "News" (Guid 8c88f208-6c77-4bdb-86a0-0c47b4316588).  

Before I could make this change I worked with Microsoft support / product to ensure that both controls are returned from the API (_api/web/GetClientSideWebParts).  Previously only NewsReel was being returned meaning "News" could never be found in that collection.

@jansenbe & @PaoloPia - I did also add a new enum value "News" so the next version of the PnP Provisioning Schema can support this web part type without being invalid.
https://github.com/SharePoint/PnP-Provisioning-Schema/pull/427


jansenbe:
Nice catch @bogeorge , thanks for the detailed research you did for this PR (after having read the related issues). Will process this one.

bogeorge:
Thanks @jansenbe it really took some digging to figure out what the heck was going on with News.  

For me what finally helped me really figure things out was the querystring ?maintenancemode=true that showed me the different id's before and after for what I *thought* was the same web part.

PR #2422, opened 2019-10-01
Fix issues reported by static analysis

jackpoz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | none

This PR fixes some issues reported by static analysis, in particular:
- Using only a sub-set of class members in GetHashCode() functions
- Using & instead of && in logical expression
- Validating the same parameter twice instead of 2 different ones
- Ignoring contructor parameter, using the class field instead
- Checking if "this" is null instead of the class member


jansenbe:
Thanks @jackpoz , this fixes a whole set of potential issues!

PR #2425, opened 2019-10-02
Improvment: Exclude "Note-Field" linked to "Taxonomie-Field" from Extract for SPO

czullu:
… - no need to include in Export.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| New sample?      | no
| Related issues?  | 

As Sp-Online creates the Note-Field by it self when we create a Taxonomie-Field, we suggest as an improvement to exclude the Note-Field linked to a Taxonomie-Field from Extract.


PaoloPia:
Hi @czullu,
Thanks for your contribution. However, I don't see an actual benefit of removing the note fields. Do you have any real use case that makes this change worth it, aside from saving on the output XML file? While waiting for your reply, I will close this item. Based on your feedback we can eventually open it again and merge it. Thanks.

czullu:
Hi @PaoloPia,
i see, i should have spend more time on description. We did need the fix in case you want to reapply a pnp-template to a list where the Taxonomie-Fields have been created by UI or Code but NoteField was not explicit set. The Apply of the PnP-Template will fail because the NoteFields beloging to the TaxonomieField have a different Id. If those NoteFields are removed from the Template and we let Sp handle it, then the Reapply of PnP Template works also for this those Lists.

I hope the reason & motivation for this fix are now more clear. A other solution would have been to add a parameter telling we want to reapply and therefore handle this case on the apply and not the export side. 

Let me know what you think about it - leave it, accept it or redo for apply?
Thanks

PaoloPia:
Hi @czullu,
Thanks for the clarification. However, I don't see any exception when applying the same template (with note fields in it) multiple times. The use case where you create the fields with UI or code and later on you want to update them through a provisioning template is a bit of an edge case. I hope you don't mind, but I'd like to not change the engine to support that edge case. I hope you will understand. Thanks.

SchauDK:
We have also removed the Note fields from our templates.
I believe at some point we discovered an issue creating Taxonomy fields if the Note field existed in the template. Can't remember what it was though.

Maybe it isn't an issue now, but I still like to have "clean" templates ;-)

PR #2431, opened 2019-10-05
Feature: Added support to overwrite tenant themes

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | NA

This PR adds support for the Overwrite attribute while processing tenant themes.
With this attribute present, we will check if a theme is present or not.

If it is not present, we will add the tenant theme.
If it is present and if its set to Overwrite, then we will update the tenant theme else skip its processing.


erwinvanhunen:
thank you @gautamdsheth !

PR #2436, opened 2019-10-08
Fix Provisioning ContentType on Folder and set as DocumentSet

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

@erwinvanhunen  As schema supports now ContentType on Folders this addition takes care that this new attribute is respected and if the ContentType is based on DocumentSet we set properties make it behave and look as DocumentSet.


erwinvanhunen:
Thank you @czullu !

PR #2437, opened 2019-10-08
Provisioing: Make sure we ExecuteQuery for newItem when no Attachments exist

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Move the check if ExecuteQuery is needed outside the Block for Update the Attachments in order to have it executed even when no Attachments exist.


erwinvanhunen:
good one!

PR #2456, opened 2019-10-18
Fix if IncludeAttachments=true but no ViewFields specified - then only attachments are extracted

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

@erwinvanhunen Since we load only all List Fields for Export if the List is Empty but add the "Attachments" to ViewFields before if IncludeAttachments is true the behavior of the Export Function in respect of the other Fields is changed. 
In this change i do queryConfig.ViewFields.Add("Attachments"); only if the ViewFields have been specified but this Field is mising.


erwinvanhunen:
that's a good one. This fixed actually showed another issue while exporting, which I will fix straight after merging this one.

Issue #2413, opened 2019-09-22
Provision template default values for unspecified attributes 


jensotto:
#### Category
[x] Bug
[ ] Enhancement

#### Environment
[x] Office 365 / SharePoint Online
[x] SharePoint 2019
[x] SharePoint 2016
[x] SharePoint 2013

If SharePoint on-premises, what's exact CU version: 
All

#### Expected or Desired Behavior
When a provisioning template is applied that does not specify one or more attributes, the value of those attributes should not be set to the hard coded default value specified in the provisining engine/template.

#### Observed Behavior
Many attributes that are not specified in the template are set to a template hard coded default value or cleared/set to an empty value.

#### Steps to Reproduce
Not specifying the NoCrawl attribute will set the site NoCrawl property to false.
Not specifying the description attribute for a list will clear any existing description for that list.

These are just examples. Almost all of the template attributes behaves in this maner, and makes it difficult to use provisioning templates for much else than provisioning new unused sites. Modifying existing sites using templates can result in lost data, or unwanted/unintended changes.

PaoloPia:
Hi @jensotto,
Thanks for your contribution. We're processing this issue and the PRs related to this, so I will close this item.

PR #2500, opened 2019-11-28
Make properties optional

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes?
| New feature?    | no?
| New sample?      | no?
| Related issues?  | #2413 

Enables provisioning engine to skip setting values where optional values are not specified in the template. Previous implementation would overwrite data even if it was not specified in the template. This makes it difficult or impossible to push out changes to existing sites using the provisioning engine.

There are also related changes in PR [#466](https://github.com/SharePoint/PnP-Provisioning-Schema/pull/466) for the provisioning schema. This is not required to be in place for this PR, but it removes the default value for many of the properties and makes some properties optional.

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Framework/Functional/Validators/ListInstanceValidator.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/Functional/Validators/ListInstanceValidator.cs
@@ -56,7 +56,7 @@ namespace OfficeDevPnP.Core.Tests.Framework.Functional.Validators
             foreach (ListInstance list in sourceCollection)
             {
                 // don't add hidden lists since they're not exported again...
-                if (!list.Hidden)
+                if (list.Hidden.HasValue && !list.Hidden.Value)
                 {
                     ProvisioningTemplate pt = new ProvisioningTemplate();
                     pt.Lists.Add(list);
--- a/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectTermGroupsTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Framework/ObjectHandlers/ObjectTermGroupsTests.cs
@@ -163,7 +163,7 @@ namespace OfficeDevPnP.Core.Tests.Framework.ObjectHandlers
 
                 Assert.IsInstanceOfType(group, typeof(Microsoft.SharePoint.Client.Taxonomy.TermGroup));
                 Assert.IsInstanceOfType(set, typeof(Microsoft.SharePoint.Client.Taxonomy.TermSet));
-                Assert.IsTrue(set.Terms.Count == 2);
+                Assert.AreEqual(2, set.Terms.Count);
 
 
                 var creationInfo = new ProvisioningTemplateCreationInformation(ctx.Web) { BaseTemplate = ctx.Web.GetBaseTemplate() };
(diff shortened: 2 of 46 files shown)


PaoloPia:
Hi @jensotto,
We had a look at your PR. First of all, you did great and huge work. Thanks! We will keep this item on hold for a while because we would like to fully and deeply test it, being the fact that the PR touches a lot of fundamental files of the provisioning engine. Secondly, we will be able to merge it once we will have the updated schema (v-next) in place. So, thanks again and stay tuned.

jensotto:
Thank you. I just want to emphasize that a schema change is NOT required for this PR (unless you do not want to introduce these changes to existing schemas, but that would require maintaining multiple versions of the engine and models). The only impact on existing schemas that I can think of is if someone has modified a schema by hand and removed some optional properties and relies on the engine to apply the default values. I don't think this is very likely, and if someone has done this (like I have), the normal and only logical expectation is that omitted properties should not change existing SharePoint configuration. 
The only benefit I can see with the schema changes is to enable making some of the properties optional that now are required. It will also remove the default values for optional properties, but because of the changes made in the provisioning engine, the engine will ignore default values if that property maps to a nullable property in the model. This way most of the new functionality will work with both existing schema and my updated schema in the PR.

I am running a custom build of PnPPowerShell that includes this PR and some other PRs opened by me (but not the schema changes). Provisioning works fine with these changes and it works as intended.
I accidentally ran one of my templates with the currently released provisioning engine (not containing this PR), and a lot of changes where made to sites that this PR will prevent. It therefore shows that this PR should work just fine without any schema changes.

On the other hand I do understand that you want to thoroughly test this change as it does make some fundamental changes in the provisioning engine as you point out.
A concern that I've had while developing this is how to prevent new contributions to introduce new functionality that does not take into account optional properties. What I mean is if a new property is implemented or introduced there is nothing to prevent the developer to continue with the previous approach that set a specific value as default if it was not specified in the template. This can happen if a new or existing property with a default value is mapped to a property in the model that is not nullable. Then the default value will be honored by the provisioning engine and put into the model. In some situation this is OK. Take for example the Overwrite property. This is an action and controls how the engine will perform it's tasks. It's also not a value that is stored in SharePoint. It therefore makes sense to have a default value for attributes like this.
I guess maintaining this potential for drift must be left to code reviewers and code maintainers to spot and prevent accepting PRs if identified. Maybe there should be some guidelines created to prevent this from getting into the code in the future? Are there some sort of code guidelines created for this and the rest of the PnP projects?

PaoloPia:
Hi @jensotto,
Thanks again for your contribution. After a deep analysis of your PR, we realized that it is a bit dangerous to merge it without proper testing and there could also be some backward compatibility issues. Moreover, we are cooking something new :-) in the fields of the provisioning engine, and it will include coding style guidelines, too. As such, we decided to keep this proposal on hold and we will make it happen in the upcoming months, together with a new major release of the engine. Thanks again for your patience and comprehension!

PaoloPia:
Hi @jensotto,
I'm sorry to inform you that we decided to not merge this PR, because it is too complex to apply the proposed changes with the risk of many issues with backward compatibility. At the same time, as I announced here in the past, we're working on a new PnP Core SDK library, where we are going to migrate the PnP Provisioning Engine (not very soon, but it will happen in the upcoming months ...). As such, we will keep your feedback into account in that new library/engine. Thanks a lot for your effort! I will close this item.

PR #2525, opened 2020-01-07
fix User Deleted Templates Folder in SitePages on Export

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix handles the situation were the user deleted the Template Folder in SitePages and Property vti_TemplatesFolderGuid contains therefore the Folder UniqueId of an not existing folder. 

If we can not resolve Folder by it's uniqueid we fallback to fix the default Foldername.


jansenbe:
Just hit this myself, so this was a timely PR :-) Thanks!

PR #2533, opened 2020-01-20
Added option to set requesting access to a web to the default owners

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no 
| Related issues?  | N/A

Added `EnableRequestAccess` method without providing e-mail addresses to instruct setting it back to the default owners of the site. After setting it to the default, trying to set it to a specific e-mail address again will set the e-mail address but will leave the selected option at the default owners in the web UI. To mitigate this, we always need to set `SetUseAccessRequestDefaultAndUpdate(false)` when setting `RequestAccessEmail`


jansenbe:
Thx @KoenZomers

PR #2534, opened 2020-01-20
Only allow one e-mail address to receive access requests

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no 
| New sample?      | no
| Related issues?  | N/A

As far as I can judge, you can only set one e-mail address to receive the access request e-mails. Updating the code to mark the current methods accepting multiple e-mail addresses as obsolete and adding new methods making it clear that only e-mail address can be set. Test it for yourself by providing multiple e-mail addresses either through the current code or the web UI. It will result in an error no matter if you use a comma or semicolon to separate multiple addresses.


jansenbe:
Thx @KoenZomers . A quick check in the code indeed shows no use of multiple email addresses, the length of the field is limited to 255.

KoenZomers:
Awesome @jansenbe! Good to have this confirmed from the source code :)

PR #2550, opened 2020-02-21
Enabled uploading document set default documents on no script sites

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Uploading default documents for document set content types are supported on noscript sites (at least it works both in Online and SP2019).


jensotto:
I might add that for some reason it's not possible to delete default documents from document sets. This is true for both GUI and CSOM

jansenbe:
@jensotto : If one can't delete them anymore is adding support to add them then useful?

jensotto:
@jansenbe Yes, it can be deleted in sites that are not noscript enabled.

jensotto:
@jansenbe As long as it is supported I don't see why it should'nt be allowed to do using the provisioning engine.

PR #2565, opened 2020-03-03
Add ability to exclude fields from content type hub when extracting provisioning template

NicolajHedeager:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | mentioned in #1222 

This PR adds the ability to exclude fields that came from content type hub. For backwards compatibility, fields are included by default.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/TaxonomyExtensions.cs
@@ -364,6 +364,20 @@ namespace Microsoft.SharePoint.Client
             return termStore;
         }
 
+        /// <summary>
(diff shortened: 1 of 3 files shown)


jansenbe:
Thanks for this change @NicolajHedeager 

We typically try to avoid additional settings in ProvisioningTemplateCreationInformation, also would then require an additional parameter in PS. @erwinvanhunen : what do you think?

jensotto:
I think this could be done by using the existing IncludeContentTypesFromSyndication setting. If content types from syndication is not included then fields probably also should not be included.

jensotto:
Also if not exporting fields from syndication it would not make sense to export content types from syndication as those content types would depend on the fields from syndication.

PR #2566, opened 2020-03-04
Extend TokenParser with ParseStringSPGroupName to ensure valid SharePoint Group name

NicolajHedeager:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

This PR ensures that the group name is always valid when provisioning a new SharePoint group.
It fixes issues where a token in Group Title prop could be parsed to a invalid character for a SharePoint Group title.

Example:

Site name: PnP/Sites/Core
Group title: "{sitename} Approvers"
parsedGroupTitle: "PnP/Sites/Core Approvers"

This will throw an error when the group is created.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectSiteSecurity.cs
@@ -46,9 +46,9 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
 
                 if (web.EnsureProperty(w => w.HasUniqueRoleAssignments))
                 {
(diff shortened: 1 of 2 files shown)


jansenbe:
Thanks for this change @NicolajHedeager. 

Question: both {sitename} and {sitetitle} tokens are resolving to the site's title. Having special chars in the site title indeed can happen, so your use case is valid. I however rather create new token called `SiteTitleForGroupToken` to implement this. Having the `ParseStringSPGroupName` feels a bit odd to me. Can you change this PR?

NicolajHedeager:
@jansenbe I was'nt sure where to implement it, but I think your suggestion sounds cleaner. 
I will have a look at using a token for it and update the PR.

jansenbe:
This should have been fixed by #2617

NicolajHedeager:
Fixed in #2617

PR #2570, opened 2020-03-05
Added list content type id token to support specifying list content t…

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | N/A

Added new token for specifying content type id for content type added to a list. This is required for specifying content types in the NewDocumentTemplates node for the default view.
The token can be used with both content type name or direct parent content type id.

Also changed name for AddListTokens method to better communicate what is being done in the method. It's not just adding tokens, it's also removing tokens. I think Rebuild is therefore better than Add. The method is also internal, so changes will have no impact on external code. This should be a safe change to do and makes the code cleaner.


jansenbe:
Thx @jensotto !

PR #2574, opened 2020-03-06
Fixing broken API call for list webhook subscriptions.

gzukula:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Fixing broken API call for list webhook subscriptions:
- "missing "guid" prefix before list id.
For a correct construction of webhook api call see: https://docs.microsoft.com/en-us/sharepoint/dev/sp-add-ins/working-with-lists-and-list-items-with-rest.
Also it seems Sharepoint documention is either not up-to-date or incomplete ("guid" is not mentioned in their guide here: https://docs.microsoft.com/en-us/sharepoint/dev/apis/webhooks/webhooks-reference-implementation

Ensuring the URL of the list is correct, and is not taken from its context (which could hold an URL of a site collection or of a website which is higher up in the hierarchy):
- When creating a subsite in the subsite, and reusing the same ClientContext, the URL was not correct and would refer to the top site collection. So it would miss the entire path of the website, where the list is located (also would contain double slashes in the path). The fix ensures that the actual URL of the list is being used in API call.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/ListExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/ListExtensions.cs
@@ -164,15 +164,15 @@ namespace Microsoft.SharePoint.Client
             // Get the access from the client context if not specified.
             accessToken = accessToken ?? list.Context.GetAccessToken();
 
-            // Ensure the list Id is known
-            Guid listId = list.EnsureProperty(l => l.Id);
+            // Ensure the list Id and Url are known
+            list.EnsureProperties(l => l.Id, l => l.ParentWeb, l => l.ParentWeb.Url);
 
             try
             {
-                return WebhookUtility.AddWebhookSubscriptionAsync(list.Context.Url,
+                return WebhookUtility.AddWebhookSubscriptionAsync(list.ParentWeb.Url,
                                WebHookResourceType.List, accessToken, list.Context as ClientContext, new WebhookSubscription()
                                {
-                                   Resource = listId.ToString(),
+                                   Resource = list.Id.ToString(),
                                    ExpirationDateTime = expirationDate,
                                    NotificationUrl = notificationUrl,
                                    ClientState = clientState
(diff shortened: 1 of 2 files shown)


jansenbe:
Thanks for this change @gzukula .

We normally only accept changes against our dev branch, but I'll make an exception for this one as I'll merge it manually. You'll be credited in the release notes, the PR will however be closed by me due to manual merge.

About the actual changes: the guid is a good catch, will add that. For the context I think you need to ensure to create a new context for the sub site you've created.

jansenbe:
closing per above discussion

gzukula:
With regards to a 2nd part of the fix, I slightly disagree that one should create and always use a context that is explicitly created for a single site only. One one hand I wonder if there are any indication from CSOM authors that indeed this is not a correct usage?

OTOH, I'm looking at similar place in webhook management code in PnP - the update API. It is also expecting that the context that is passed is linked to the exact site, as the context's URL is then used to invoke the REST call. However, here is a typical scenario when the subscriptions are updated:
- Upon list modification event the "siteUrl" field contains a site-collection URL and not the subsite URL for which the webhook was created (see https://docs.microsoft.com/en-us/sharepoint/dev/apis/webhooks/overview-sharepoint-webhooks). The documentation is not very obvious and implies that the actual website's URL should be used, however in reality it's the site collection one (maybe because having deeper hierarchies in Sharepoint is not too common).
- So when it comes to extending a subscription, one gets a handle to a web by using context linked to a site collection. Then the same context should be used to get a name of that website (remember, it's only website's GUID that is passed in the original message), and to create yet another context. I would say it's at least inefficient (as requires multiple roundtrips to Sharepoint) to be forced to create another context, merely to invoke a single REST API. 

Luckily PnP provides another and more specialized public API (used internally by ListExtensions) where one can explicitly control the REST endpoint - WebhookUtility.UpdateWebhookSubscriptionAsync. So I would say it's a matter of taste which one should be used, so we can use that one, and won't need to create a local copy outside of PnP :)

In the end hopefully CSOM shall be extended to include such API on its own, so we wouldn't need to argue about the correct of CSOM context.

PR #2579, opened 2020-03-10
Fix encoded first { in fileuniqueid token dedection in navigationlink

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fileuniqueid token in Navigationlink not dedected if Link starts with "%7b{" and not "{{"

Fix Regex in order to handle encoded first { as posible option in token dedection


jansenbe:
Thx @czullu

PR #2582, opened 2020-03-13
Fixed incorrect implementation of content type id best match

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Previous implementation of content type id BestMatch was incorrect. It did not check for a common parent and could therefore end up with not finding av best match even though a content type sharing the same parent existed. For example if an Item list content type on a list with id 0x010100C284F907FFBF7D45B2B102E254551281 and an Item list content type on a different list with id 0x010100CE36B052C59F4C46ADD8ADA2F8668F3D the previous implementation would not find any match. It should identify 0x010100CE36B052C59F4C46ADD8ADA2F8668F3D as the best match for 0x010100C284F907FFBF7D45B2B102E254551281.

Implementation of BestMatch in this PR has been taken from the server side implementation of the same functionality.


jensotto:
@PaoloPia @jansenbe Can you please complete this PR before this project is shut down. Will this be included in PnP Framework?

PR #2584, opened 2020-03-13
Fixed changing view title casing

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Changing the case of the view title will not update its value. Workaround for this is to change the title to a temporary value before setting the view title to the correct value.

This can be tested when there is a veiw with url "{site}/Library/Forms/A view title.aspx" and Title "A View Title"

When the template is applied multiple times a new view will be created each time with the name specified for the aspx file in the url attribute.


jansenbe:
Thanks for this pr @jensotto , it's an edge case scenario but nevertheless it's good to get it in, this will make the provisioning engine better.

jansenbe:
fyi: your pr was for master but I merged into dev, so pr will be marked as merged later today due to the master merge process

jensotto:
sorry about that.

PR #2585, opened 2020-03-14
Fixed IsChildOfDocumentSetContentType in ContentType Provisioning

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no 
| Related issues?  | N/A

Microsoft.SharePoint.Client.DocumentSet.DocumentSetTemplate.IsChildOfDocumentSetContentType requires an ExecuteQuery before it actually does something and returns a valid value. At present it would always return false, regardless if the content type did inherit from the document set content type. Confirmed by using Fiddler to look at the network traffic behind the execution.


jansenbe:
thx @KoenZomers

PR #2594, opened 2020-03-20
Retrieve all groups from Graph endpoint

ohaak:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no 
| New sample?      | no 
| Related issues?  | 

Currently the method UnfiedGroupsUtility.ListUnfiedGroups will always return a maximum auf 999 groups from the Microsoft Graph endpoint, despite the fact that it is basically indendet to spare the calling method the effort to retrieve all the groups in page chunks, all the while handling with start- and endIndex value.

Reason for this behaviour is a probably unwanted mingling of endIndex and the actual pageSize used for the Graph call. "endIndex" defaults to 999 which is currently the allowed maximum page size to retrieve on chunk of groups from the endpoint. Increasing the value for "endIndex" will result in an error stating that a value greater than 999 is invalid.

So increasing the default value for endIndex to be able to retrieve more than 999 won't do any good, because we will just get an exception. Leaving it to 999 will operate the MS Graph call with the correct maximum page size, but internal follow-up calls will be prevented by terminating the overall processing as soon as the actual group count reaches the endIndex value. So currently we can manage to get any lesser number than 999 groups from the backend, but no more.

Decoupling pageSize (new parameter in method signature) and endIndex will allow us to operate with a customizabel pageSize retrieving the groups from Graph (defaults to 999) and not set this value per default as endIndex as well. So specifying a bigger endIndex is now possible.

Best thing however is just leaving the default values with no endIndex value defined. This way method will return all groups from the Graph backend which is pretty much what the method documentation (and the PowerShell command based on it) state they would do.


jansenbe:
Good fix @ohaak , but we can't change public method signatures. You can rework this PR with a new public method with the correct parameters and mark the current one as obsolete.

ohaak:
@jansenbe I've reworked the pull request as requested - sorry for the delay. I hope it's ok now. Wasn't quite sure how to proceed with the method name. Leaving it "as-is" for the new method would probably have been the best option, but that way it's a little bit fuzzy to comment the change in the deprecated method attribute on the old one.

Issue #2595, opened 2020-03-23
Provisioning Template.  Omitting Sharing Settings results is default being Disabled


SteveClements:
#### Category
[ X ] Bug
[ ] Enhancement

#### Environment
[ X ] Office 365 / SharePoint Online
[ ] SharePoint 2016
[ ] SharePoint 2013

Provisioning Schema: 202002

#### Description
When using the Provisioning Template, I'm using Tenant Template, probably applies to any as it's related to Tenant config.
If you don't specify `<pnp:SharingSettings>` in the XML file, the type resolver sets the default value to 'Disabled' rather than null.

#### Expected or Desired Behavior
Sharing Settings should not be changed at all if `SharingSettings `is not specified in the template.

#### Observed Behavior
All Sharing Settings in the Tenant are removed and set to Disabled.

#### Steps to Reproduce
1. Ensure sharing settings in the Tenant are not Disabled.
2. Create a tenant template with a `<pnp:Tenant>` section
3. Do not specify `SharingSettings`
4. Run the template
5. Check sharing settings in the tenant, they will be Disabled.

#### Fix
I've fixed this is a local build of pnp core by changing
`OfficeDevPnP.Core.Framework.Provisioning.Providers.Xml.Resolvers.V202002.SharingSettingsFromSchemaToModel`

Resolve will return null if `SharingSettings `are null, rather than a default instance of `Model.SharingSettings`, which has an `enum `where `0 = Disabled`.

I will try and create a clean branch and submit a PR.  But in the meantime, are there any potential issues with this fix?  It seems to be expected as 
`TenantHelper.ProcessSharingSettings` is checking for `null `and doing nothing (as expected) if it is.

jansenbe:
change was merged

PR #2596, opened 2020-03-23
Resolves issue where sharing settings defaults to disbled even when S…

SteveClements:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

As per item #2595 a simple change to `OfficeDevPnP.Core.Framework.Provisioning.Providers.Xml.Resolvers.V202002.SharingSettingsFromSchemaToModel` to return `null` when `SharingSettings` are not used in the template XML.
Currently it will return a default value which is `Disabled`


jansenbe:
Thx @SteveClements , good fix!

PR #2602, opened 2020-03-26
Fix: ensuring correct usage of EnsureProperties method

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | NA

In this PR, replaced the `EnsureProperties` method with `EnsureProperty` method across the framework where necessary.


jansenbe:
Thanks for this change @gautamdsheth. I'm wondering what additional value you see with this change, I don't think EnsureProperty is more performing and since the return variable is never used there's no functional reason for changing this.

gautamdsheth:
Agree, mainly its for consistency across the solution.
We did the same thing in the modernization project as well some time back.

PR #2603, opened 2020-03-30
Fix: additional check to determine whether theme exists in the tenant

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | NA

This PR adds an additional check to determine whether a particular custom theme exists in the tenant. If it exists, we will apply that theme, otherwise will show a warning and skip it. Currently, if the theme doesn't exist, the provisioning stops at this stage only


jansenbe:
Good update @gautamdsheth !

PR #2609, opened 2020-04-07
Handle special char in SiteTile beeing replaced by _ for SP-Groupnames

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

If site with title using special chars are created - you will get the Error: 
The group name is empty, or you are using one or more of the following invalid characters: " / \ [ ] : | < > + = ; , ? * ' @

Explanation: If the user creates a site with Title "there's my site" then microsoft names the SP Groups "there_s my site Owner". This leads to the issue that tokenized groups like "{sitename} Owner" will not be resolved correctly. This fix implements the same replacement logic as ms using in UI in the new ParseGroupTitleString in order to fix this issue.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/Extensions/SecurableObjectExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/Extensions/SecurableObjectExtensions.cs
@@ -31,7 +31,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.Extensions
             {
                 if (!roleAssignment.Remove)
                 {
-                    var roleAssignmentPrincipal = parser.ParseString(roleAssignment.Principal);
+                    var roleAssignmentPrincipal = parser.ParseGroupTitleString(roleAssignment.Principal, context.Web);
 
                     Principal principal = groups.FirstOrDefault(g => g.LoginName.Equals(roleAssignmentPrincipal, StringComparison.OrdinalIgnoreCase));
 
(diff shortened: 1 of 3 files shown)


jansenbe:
@czullu : it feels wrong to have this logic built into the tokenparser class (although I agree it's the easiest fix). Did you consider introducing a new token for this and when extracting groups use that one by default in the default group titles

czullu:
Hi Bert
I know what you mean, but I was looking for a fix which does not impact Templates which Customers did already build.
However – if we create a new Token then I could take care of the correction in a prepare Task before apply. Any suggestion on the Tokenname? How about {sitenameencoded}?

With best regards
Christian

Von: Bert Jansen <notifications@github.com>
Gesendet: Donnerstag, 9. April 2020 10:56
An: pnp/PnP-Sites-Core <PnP-Sites-Core@noreply.github.com>
Cc: Christian Zuellig <czullu@outlook.com>; Mention <mention@noreply.github.com>
Betreff: Re: [pnp/PnP-Sites-Core] Handle special char in SiteTile beeing replaced by _ for SP-Groupnames (#2609)


@czullu<https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fczullu&data=02%7C01%7C%7Ce107f3d409de448c9b5f08d7dc63e299%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637220193862921677&sdata=lbc972nwXOTEf%2F023ClYrt7SlJTZxfb0IFlqJ6hEf8E%3D&reserved=0> : it feels wrong to have this logic built into the tokenparser class (although I agree it's the easiest fix). Did you consider introducing a new token for this and when extracting groups use that one by default in the default group titles

jansenbe:
Since the encoding is specific for groups I would make that explicit: call the token GroupSiteTitle and have it listen to {groupsitetitle} and {groupsitename}

jansenbe:
Closing this one as it's replaced by #2617

PR #2610, opened 2020-04-07
Have to set file properties before checkin/publish.

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Error: Files with properties in Libraries with ForceCheckout on failed to be provisioned.

Fix: The function SetFileProperties has to be called before checkin/publish in ObjectFiles


jansenbe:
Good find @czullu

PR #2615, opened 2020-04-10
Revert "Fix: ensuring correct usage of EnsureProperties method"

jansenbe:
Reverts pnp/PnP-Sites-Core#2602 

running extraction of template showed this code to break:
 var iprFeature = features.GetById(featureID);
            iprFeature.EnsureProperty(f => f.DefinitionId);

Given we're releasing today, I'm reverting #2602


jansenbe:
@gautamdsheth : FYI, I'm reverting your PR as it's causing unforeseen issues. I don't have the bandwidth to test each individual change, so I'm taking the safe route here and revert this pr as it's anyhow not adding funtional improvements or bug fixes

PR #2617, opened 2020-04-13
Handle special char in SiteTile beeing replaced by _ for SP-Groupnames

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | yes
| Related issues?  | mentioned in #2609

What's in this Pull Request?
as agreed this fix implements 2 new tokens groupsitetitle and groupsitename and makes use of this new Tokens while extract the Site Groups and Roles. This tokens represent the sitetitle with special char " / \ [ ] : | < > + = ; , ? * ' @ replaced with _.


jansenbe:
Nice work @czullu , thanks for doing this!

PR #2634, opened 2020-04-29
Fixing NullReferenceException in case the authentication with credentials fails

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Fixing NullReferenceException in case the authentication with credentials fails, such as in the scenario where a user enters incorrect credentials. Returning NULL instead to indicate we don't have an AccessToken.


jansenbe:
thx @KoenZomers

PR #2638, opened 2020-05-06
Remove tokens from TokenDictionary and ListTokenDictionary when rebui…

ZeroneBit:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

We should remove tokens from TokenDictionary and ListTokenDictionary when rebuilding list tokens in the function RebuildListTokens. As there may be redundant tokens in the dictionary.

#### Steps to Reproduce
The issue can be reproduced by the following steps.

1. Use GetProvisioningTemplate to create a template at web level (change the scope to Web) which contains,

- An Events webpart on home page (connected to the events list).
- An events list in this web.

2. Use ApplyProvisioningTemplate to apply this template to the root site of a new site collection.
3. Use ApplyProvisioningTemplate to apply this template to the sub-site of the site collection.

The Events webpart at the sub-site level doesn't connect to the events list.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/TokenParser.cs
@@ -646,11 +646,22 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
         {
             web.EnsureProperties(w => w.ServerRelativeUrl, w => w.Language);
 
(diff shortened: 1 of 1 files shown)


jansenbe:
Thanks for this pr @ZeroneBit . Would it be possible to share your sample template as well?

ZeroneBit:
@jansenbe here is the xml template. 
[SiteTemplate.zip](https://github.com/pnp/PnP-Sites-Core/files/4612780/SiteTemplate.zip)

Please make sure there is no Events list in the web before provisioning.

PR #2642, opened 2020-05-11
Bugfix in paging for retrieving AAD Users

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no 
| New sample?      | no
| Related issues?  | N/A

Bugfix in paging when requesting Azure Active Directory users and minor help text fix. endIndex argument was used to measure the amount of results that would be returned instead of the index of the last item that should be returned.


KoenZomers:
Canceling. Will be replaced by https://github.com/pnp/PnP-Sites-Core/pull/2643

PR #2659, opened 2020-05-29
Fix ExecuteQuery implementation failing when Digest does expire while retry

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

If the Digest does expire while we in Retry we end up with uncommited changes or properties not loaded but we not throw an error. The Fix does check lifetime of Digest before Query Execution and get Digest refreshed if near to expire before start QueryExecute.


jansenbe:
Hi @czullu , how often are you facing this issue? I'm not immediately merging this one as this type of check should be ideally in the CSOM code base + it's on overhead (reflection) on every executequery. We'll need to further discuss this

czullu:
Hi Bert
Totally agree that it would be best handled inside CSOM. It did happen quite often at the beginning of corona slowdown of O365/SPO when retry was more often happening.
Usually a PnP Export of a Site would fail at a random stage with a message that a property was not loaded.
Debugging this issue we figured that the digest did expire while in ExecuteQueryRetry loop – what happens then is, we get a new ClientContext and as this Context does not have any information of the outstanding Loads and therefore will successful execute and return to code. In this way we end up with a Context not having loaded the Data we thought.

As we have this fix in out PnP Build we never had the issue anymore – so I’m not really in the position to answer the question how often this still happens.

With best regards
Christian

Von: Bert Jansen <notifications@github.com>
Gesendet: Donnerstag, 4. Juni 2020 15:40
An: pnp/PnP-Sites-Core <PnP-Sites-Core@noreply.github.com>
Cc: Christian Zuellig <czullu@outlook.com>; Mention <mention@noreply.github.com>
Betreff: Re: [pnp/PnP-Sites-Core] Fix ExecuteQuery implementation failing when Digest does expire while retry (#2659)


Hi @czullu<https://eur06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fczullu&data=02%7C01%7C%7C928612f8e814418f700208d8088ccdd8%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637268748118139387&sdata=sRvBA%2Fu9%2BBphdGw39Oc2PSo%2FapfijeQm6W%2FGj8qoFY8%3D&reserved=0> , how often are you facing this issue? I'm not immediately merging this one as this type of check should be ideally in the CSOM code base + it's on overhead (reflection) on every executequery. We'll need to further discuss this

jansenbe:
Good to hear this type of fix really helped with your stability. Another possible way to prevent this is not using classic (legacy) username/password auth as that depends on this, access token based auth does not use the form digest and should also help here.

czullu:
Hi Bert, we use a Azure App (AppId & Cert) plus userimpersonation-token which seams to run into that issue as well. As we can do much App-Only - there are some things we need to do on-behalf.

jansenbe:
@czullu : that's shouldn't be the case. Can you try with explicetely setting FormDigestHandlingEnabled to false on the contexts created like that

jansenbe:
Hi @czullu , I'm going to close this one because for SPO the recommended approach is using access token based flows (legacy auth is off on many tenants already) and in those cases the formdigest is not relevant.

PR #2660, opened 2020-05-29
Fix CreateFolder leads to version count equal 2 instead of 1

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

The current process does create the Folder and the set the Properties - therefore the Folder Version will be 2 instead of 1 as expected. Still not understand why this matters, but fixed it by doing the FolderItem update as SystemUpdate.


gautamdsheth:
Hello Christian, nice to meet you !
Can we use `UpdateOverwriteVersion()` instead of `SystemUpdate()` ? The `SystemUpdate` method is not available for SP 2016 and 2013

czullu:
Hi, yes, sure can change that to night (aprx in 6h).
Thanks for the input, working mostly with spo, so did not think about it.
Best regards
Christian

Freundliche Grüsse
Christian

czullu:
@gautamdsheth Hi, change is committed.

PR #2663, opened 2020-05-29
Add the possibility to have ContentTypes updated on existing Lists

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

Added the Option UpdateContentTypeOnExistingLists to ProvisioningTemplateApplyingInformation in order to tell UpdateList that it can Update ContentTypes on existing Lists even the user can not select CT him self.

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -116,7 +116,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers
                             {
                                 scope.LogDebug(CoreResources.Provisioning_ObjectHandlers_ListInstances_Updating_list__0_, templateList.Title);
                                 var existingList = web.Lists[index];
-                                var returnTuple = UpdateList(web, existingList, templateList, parser, scope, isNoScriptSite);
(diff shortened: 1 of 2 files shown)


jansenbe:
Thanks @czullu , I've assigned @erwinvanhunen to this PR

Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
```diff
@@ -1467,7 +1467,8 @@ private static XElement PrepareField(XElement fieldElement)
                     isDirty = false;
                 }
 
-                if (existingList.ContentTypesEnabled)
+                //it might make sense to update ContentType on existing list in case we want to add/refresh a ContentType even if users should not choose
+                if (existingList.ContentTypesEnabled|| UpdateContentTypeOnExistingLists)
```
heinrich-ulbricht, line 1514: Maybe the new property shouldn't be called `UpdateContentTypeOnExistingLists` but something like `ForceUpdateListContentType` or so. `UpdateContentTypeOnExistingLists` sounds like I have to to always set it to `true` because otherwise CTs won't be updated. But this is not the case. They _are_ updated, but only if CTs are enabled on the list.
Or what about `UpdateContentTypesOnListsWithDisabledContentTypes`?

Naming is hard :-/

czullu: Hi- it could be "ForceUpdateListCTWhenManageCTDisabled" - but maybe a better comment with the parameter would be sufficient as well.

PR #2680, opened 2020-06-15
Fix for being able to apply custom themes from provisioning templates

### LINKED ISSUE #1801: Unable to Provision Modern Team Site Theme and Logo using Apply-PnPProvisioningTemplate
Category

[x] Bug

Environment

Expected or Desired Behavior

When saving a template from a modern site which has a OOB theme and Custom Logo on the page , and applying the template, I expect the Theme and Logo added to show.

Observed Behavior

When applying a template saved with Theme and Logo, the Theme and Logo are not rendered - even though they appear in the extracted XML

Steps to Reproduce

Go to Modern Team Site
Click on Gear Icon
Choose the option "Change the look"
Go to SiteAssets and upload the New Site Logo Image

Extract a template from the site with Get-PnPProvisioningTemplate -Out .\home.xml , then run Apply-PnPProvisioningTemplate -Path .\home.xml.

The Page will be blank and the old theme only applied over the page and Logo was not changed.

cmzaragoza:
| Q               | A
| --------------- | ---
| Bug fix?        |  yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #1801 

Small change in the _ProvisioningObjects_ method from the _ObjectTheme_ class, that allows to apply a custom theme previously uploaded to the Tenant to a modern site.


jansenbe:
Hi @cmzaragoza , we've been provisioning custom themes for our lookbook sites without issues (e.g. https://github.com/SharePoint/sp-dev-provisioning-templates/blob/master/tenant/ContosoLanding/source/template.xml#L24). Did you check if that approach works for you as well? Also we don't want to require high permissions when this is not absolutely needed

cmzaragoza:
Hi @jansenbe , the thing is that you're able to apply a custom theme to a site only by specifying the colour palette. That can be enough for many scenarios, but what happen if we need to apply a custom theme to , let's say, 10 site collections. Having to repeat the colour palette might introduce mistakes, and also doesn't allow you splitting the templates. So the idea is being able to provisioning a custom theme at tenant scope (something that PnP Core allows already), and then being able to reuse it for each site collection. That's something you can do easily with the PnP PowerShell, but is not covered in the PnP Core, hence this pull request. Regarding the high permissions, I think it only requires the same kind of permissions as when you apply a provisioning template at tenant level (eg: Apply-PnPTenantTemplate).

heinrich-ulbricht:
If the provisioning account hasn't got enough permissions we could log a warning and skip this step. Ideally the warning contains a hint to the required permissions. This would be a very pleasant user experience in my view :)

cmzaragoza:
Hi @heinrich-ulbricht. Thanks for the suggestion. I've applied the change :)

PR #2684, opened 2020-06-22
Fix incorrect onprem eventreceiver reference

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Fixed incorrect location based metadata event receiver reference for SharePoint 2016 and 2019.
Previous version used a reference for SharePoint 2013 for SharePoint 2016 and 2019.
Assembly redirects added by default in SharePoint 2016 and 2019 installations probably made this not a real issue.


jansenbe:
Good point, indeed better to have this done the right way. Thanks for the fix @jensotto !

PR #2687, opened 2020-06-24
Adding authentication cookies for SPO Admin domain

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | discussed in https://github.com/pnp/PnP-PowerShell/issues/2742

In PnP PowerShell, when using `Connect-PnPOnline -Url https://tenant.sharepoint.com -WebLogin` and then running a PnPAdminCmdlet such as `Get-PnPTenant`, it would throw a 401 access denied as it doesn't have the FedAuth cookie for the tenant-admin.sharepoint.com domain. With this update, when connecting, it will add the FedAuth cookie for both the tenant.sharepoint.com as well as for the tenant-admin.sharepoint.com domains so the PnPAdminCmdlets will auto elevate and work correctly.

### DIFF
--- a/Core/OfficeDevPnP.Core/AuthenticationManager.cs
+++ b/Core/OfficeDevPnP.Core/AuthenticationManager.cs
@@ -425,7 +425,10 @@ namespace OfficeDevPnP.Core
                         }
                         if (authCookies != null)
                         {
-                            authCookiesContainer.SetCookies(siteUri, string.Join(",", authCookies));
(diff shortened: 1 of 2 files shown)


jansenbe:
Thanks @KoenZomers , simple and good fix for a problem that initially seemed more complex!

KoenZomers:
@jansenbe Indeed, not too bad :) Only downside of this solution is that it will not work for dvNext enhanced tenants with vanity domains. Not sure how many of those we still have around anyway. Do you have some insight into this?

PR #2688, opened 2020-06-25
Added possibility to set AllowDeletion for a field

jensotto:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | N/A

Adds the possibility to set and get the AllowDeletion for a field. This is not exposed by CSOM, but is available by manually manipulating the SchemaXML for the field.
This can be usefull to prevent users from accidentally deleting a field and loosing data contained in that field for all items in a list. Also it will make it possible to prevent deletion of a field from site columns.

A UserVoice request has been created to expose this in CSOM. In the meantime this workaround should cover it.
https://sharepoint.uservoice.com/forums/329220-sharepoint-dev-platform/suggestions/40701070-expose-field-property-setter-for-allowdeletion


jansenbe:
Thanks for this one @jensotto !

PR #2701, opened 2020-07-08
Fix: Token parsing changes

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| New sample?      | no
| Related issues?  | Not sure

This PR adds minor token parsing changes.

1) After clearing the token for group changes, added that token again for use downstream.
2) Replaced the HTTP request with the unified groups method due to 100 limit of groups retrieved.


jansenbe:
Thanks @gautamdsheth !

PR #2707, opened 2020-07-16
Allow providing a specific URL with GetWebLoginClientContext to log on to

KoenZomers:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | N/A

Added `loginRequestUri` to `GetWebLoginClientContext` in `AuthenticationManager.cs` to allow providing a specific URL to log on to. This solves an issue when you are connecting using this method to the SharePoint Online Admin Center. If the tenant you're connecting to is using the new modern experience, it will go through a redirect before it loads the page and completes the login. If the tenant you're connecting to uses the classic SharePoint Online Admin Center experience still, it will load the classic page listing all site collections which can take up to minutes to be loaded and thereby complete the login cycle. By adding this optional parameter, we can explicitly log on to a URL that is known to be fast, such as /_layouts/15/settings.aspx.

### DIFF
--- a/Core/OfficeDevPnP.Core/AuthenticationManager.cs
+++ b/Core/OfficeDevPnP.Core/AuthenticationManager.cs
@@ -21,6 +21,7 @@ using System.Net.Http;
 using Newtonsoft.Json.Linq;
 using OfficeDevPnP.Core.Utilities.Context;
 using System.Web;
+using System.Windows.Threading;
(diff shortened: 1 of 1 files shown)


jansenbe:
Good change, thanks for that @KoenZomers

PR #2724, opened 2020-08-05
Checking for and ignore missing principals when provisioning object security

### LINKED ISSUE #2723: [FEATURE] Ignore missing principals when applying the provisioning
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when moving templates between tenants. The templates often contain permissions for users that might not be present on the target tenant. But I don't care. It should just carry the existing ones over. 

Currently provisioning fails and I have to remove non-existing principals by hand. One by one. Each removal needs another provisioning run since I don't know which users will be missing. This takes time.

**Describe the solution you'd like**
I'd like the enging to ignore missing principals. Issue a warning and carry on privisioning.

**Describe alternatives you've considered**
One alternative could be to prepare the template file to not contain non-existing principals.

heinrich-ulbricht:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | fixes #2723

Unfortunately some lines of code had to be changed to achieve what I described in #2723. It looks bigger than it actually is because the diff view seems to be confused by the moved code blocks. I extracted methods for code that has already been copy and pasted and that I would've had to copy again.

Assume you have to provision a  template with lots of role assignments for lots of principals. So far they were all applied at once. If one principal couldn't be found anymore the whole operation failed and the provisioning stopped. This is bad. But this logic is still there as step 1 - if it succeeds all is well.

I added a second step that is only triggered if step 1 fails. Then each and every principal will be checked for existance. Role assignments for non-existing principals will be removed. Then the assignment of role assignments is tried again and should now succeed. Non-existing principals are logged for the user to review:

image

(What I'm not so happy about is that I needed to pass the `WriteMessage` delegate to `SetSecurity`. Maybe there is a better solution. I need this delegate to write the warnings about non-existing principals.)

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/Extensions/SecurableObjectExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/ObjectHandlers/Extensions/SecurableObjectExtensions.cs
@@ -49,7 +49,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers.Extensions
                         // catch user not found
                         if (ex.ServerErrorCode == -2146232832 && ex.ServerErrorTypeName.Equals("Microsoft.SharePoint.SPException", StringComparison.InvariantCultureIgnoreCase))
                         {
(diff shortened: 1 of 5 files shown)


jansenbe:
@heinrich-ulbricht : nice addition adding additional reliability when applying templates across tenants.

PR #2728, opened 2020-08-05
Log missing users which couldn't be added to a group and fix potential exception

heinrich-ulbricht:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| New sample?      | no
| Related issues?  | NA

Log users that couldn't be added to groups. This also makes it more consistent with other places where missing users currently often lead to provisioning being interrupted. As a PnP user I want to be informed about this. The pull request makes it look like so:

image

Also there is a potential error in the exception handler when accessing the group title. Fixing it while here.


jansenbe:
Nice improvement @heinrich-ulbricht , thanks for sharing this one with us.

PR #2730, opened 2020-08-07
Ensure decode of folderpath from uri segments to fix issue when url contains spacing

NicolajHedeager:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

Fixes issue in extraction of template files where the folder property is not tokenized if site collection url contains spacing


jansenbe:
Thanks @NicolajHedeager !

PR #2745, opened 2020-09-07
Resolve fileuniqueid or fileuniqueidencoeded in DataRow Value if not already resolved by parser

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

The fileuniqueid or fileuniqueidencoded token in a DataRow Value was not resolved if the token was not previously registered.
In the Fix i do add the UniqueID of Folders created from File Path to the Parser as well and check after the parser processed the Fieldvalue if there is still one of the fileuniqeuid tokens - if so i try to get the uniqueid.


joshtransient:
I stumbled upon this while I was trying to figure out while fileuniqueid and fileuniqueidencoded weren't parsed by the engine when I `Apply-PnPProvisioningTemplate ... -ExcludeHandlers Files` (to save bandwidth and not have to re-upload 50MB worth of assets). Will this resolve that problem?

czullu:
Hi, if you only need it for data row values than you good with that fix. The fix for navigation is as far as I remember part of the official release now and for client side pages Bert had built it from beginning.

With best regards
Christian


Freundliche Grüsse
Christian

jansenbe:
Thanks @czullu !

PR #2746, opened 2020-09-07
Fx CustomFormatter Compare to treat "" and Null as same string

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

It seams that a SiteColumn can have CustomFormater Field Value Equals "" while the depending Column on the List has Field the Value set to null. Since both mean have no CustomFormater i treat them as same and try not to preserve this as extra Info from the Column on the List.


jansenbe:
Nice catch @czullu

PR #2762, opened 2020-09-18
Handle AddWebPartToWikiPage() being called on a non-wiki page

jackpoz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes https://github.com/pnp/PnP-PowerShell/issues/2921

Currently calling AddWebPartToWikiPage() on a non-wiki page tries to run a Regex on the content of the list field "WikiField" which is null. This Pull Request adds a null check and returns null as result webpart (which is the same existing behavior if the page is not found)


gautamdsheth:
@jansenbe  - going to test this one.

PR #2774, opened 2020-09-24
Retry when inner exception for WebException is SocketException error AddressAlreadyInUse

NicolajHedeager:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | N/A

In SPO we have started to see the WebException "Unable to connect to the remote server" with an inner SocketException saying
"Only one usage of each socket address (protocol/network address/port).....".

This PR includes the specific SocketException in the existing retry logic for ExecuteQuery and retries the request if the SocketException occurs.

### DIFF
--- a/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
+++ b/Core/OfficeDevPnP.Core/Extensions/ClientContextExtensions.cs
@@ -15,6 +15,7 @@ using Newtonsoft.Json;
 using OfficeDevPnP.Core.Utilities.Async;
 using System.IdentityModel.Tokens.Jwt;
 using System.Collections.Generic;
+using System.Net.Sockets;
 using OfficeDevPnP.Core.Utilities.Context;
 using OfficeDevPnP.Core.Framework.Provisioning.ObjectHandlers;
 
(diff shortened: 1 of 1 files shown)


jansenbe:
@NicolajHedeager : think this fix is not actually fixing the root cause here. What you see is port exhaustion, when there are too many http clients created and disposed in a short time window then their ports are not freed up in time. There's guidance in .Net on how to use a HttpClient (see https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net-core) which we've not implemented in pnp sites core at the moment. The new PnP Core SDK (aka.ms/pnp/coresdk) uses a very clean http model, but unfortunately this is not yet in pnp sites core. When you these issues are you then using PnP Sites Core in an Azure function (or similar) doing a lot of requests? A check check in code revealed several places where the http client is used wrongly, so if you're using these bits "in large scale" you might run into this.

Tagging @PaoloPia  and @erwinvanhunen

jansenbe:
Hi @NicolajHedeager , I've created an issue to track this for PnP Framework as we're going to fix it over there. I'll close this PR as such.

NicolajHedeager:
@jansenbe  I think you are right. We are generating a lot of requests from a colud environment (Azure App Service), so this PR properly does'nt fix the root cause. Thanks for handling this, I will keep track of the issue in the PnP Framework.

PR #2776, opened 2020-09-25
Fix issue with handling of terms with comma and provided GUID

### LINKED ISSUE #1955: Import-PnPTaxonomy, Creates wrong hierarchy when term contains comma
#### Category
* Bug

#### Environment
* Office 365 / SharePoint Online
* SharePoint 2016
* SharePoint 2013

If SharePoint on-premises, what's exact CU version: October 2018

_Note: I know there's a separate repo for the Posh Cmdlets, but I traced the code through and I believe the issue is at two methods within `TaxonomyExtensions.cs` - `NormalizeName` (which needs to understand that 'name' might have a guid tacked onto the end) and `ImportTermSetLineParse` (which needs to do the same)_

Using an XML file from a previous run of `Export-PnPTaxonomy` with -IncludeID, the imported hierarchy matches the original hierarchy exactly with the original IDs preserved for each level in the hierarchy.

#### Observed Behavior
Where a term contains a comma, the hierarchy is broken, so a child term is created with text after the comma. There is also a double quote character at the start of one term and at the end of the child term.

#### Steps to Reproduce
Create a group/termset/term hierarchy in SharePoint using the UI, and add a term with a comma. e.g.
`My Group\My TermSet\My Term\Hello, World`

`Export-PnPTaxonomy -TermSetId {your termsite id} -Path terms.xml -IncludeID`

Then

`Import-PnPTaxonomy -Path .\termsxml`

Note the resulting hierarchy, with the weird quote characters.

`My Group\My TermSet\My Term\＂Hello\World ＂`

reusto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #1955

Fixes an issue where importing terms which include commas, while also providing a GUID for the terms, would create the wrong hierarchy in the termstore.

A import like this:
> Group;#XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX|Set;#XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX|"TermX,TermY";#ZZZZZZZZ-ZZZZ-ZZZZ-ZZZZ-ZZZZZZZZZZZZ

would create a structure like this
> Group > Set > "TermX > TermY"
> TermY" has the GUID ZZZZZZZZ-ZZZZ-ZZZZ-ZZZZ-ZZZZZZZZZZZZ

instead of
> Group > Set > TermX,TermY
> TermX,TermY has the GUID ZZZZZZZZ-ZZZZ-ZZZZ-ZZZZ-ZZZZZZZZZZZZ

### DIFF
--- a/Core/OfficeDevPnP.Core.Tests/Extensions/TaxonomyExtensionsTests.cs
+++ b/Core/OfficeDevPnP.Core.Tests/Extensions/TaxonomyExtensionsTests.cs
@@ -826,6 +826,37 @@ namespace Microsoft.SharePoint.Client.Tests
         }
 
         [TestMethod()]
+        public void HandleTermsWithCommanAndGuidTest()
+        {
+            using (var clientContext = TestCommon.CreateClientContext())
+            {
+                var site = clientContext.Site;
+
+                var termName1 = "1stComma,Comma";
+                var term1Guid = Guid.NewGuid();
+                var termName2 = "2ndComma,Comma";
+                var term2Guid = Guid.NewGuid();
+                var termName3 = "3ndComma,Comma";
+                var term3Guid = Guid.NewGuid();
+                List<string> termLines = new List<string>();
+                string termSrc1 = _termGroupName + ";#" + _termGroupId + "|" + _termSetName1 + ";#" + _termSet1Id + "|\"" + termName1 + "\";#" + term1Guid.ToString();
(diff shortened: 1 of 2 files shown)


jansenbe:
@reusto , thanks for this PR and really appreciate you took the effort to make it available in both PnP Sites Core as in PnP Framework

PR #2777, opened 2020-09-28
Fix: Removed unnecessary compiler directive and using that prevented build of PnPCore.sln

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Removed unnecessary compiler directive and using that prevented build of PnPCore.sln. The compiler directive excluded a method used in another part of the solution which was not excluded and caused the build of PnPCore.sln for .NET Standard 2.0 to fail. The previously excluded method has been manually tested from .NET Core 3.1 without any issues so there should be no need to exclude it.

### DIFF
--- a/Core/OfficeDevPnP.Core/AuthenticationManager.cs
+++ b/Core/OfficeDevPnP.Core/AuthenticationManager.cs
@@ -21,7 +21,6 @@ using System.Net.Http;
 using Newtonsoft.Json.Linq;
 using OfficeDevPnP.Core.Utilities.Context;
 using System.Web;
-using System.Windows.Threading;
 
 namespace OfficeDevPnP.Core
 {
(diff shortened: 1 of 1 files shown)


jansenbe:
Hi @patrikhellgren ,

We'll soon (end of next week probably) ship a preview version for PnP Sites Core for .NET standard, which we've branded PnP Framework. That one will be one going forward for folks needing .NET Standard, so I would recommend checking out that one ones it's available. The current PnP Sites Core version will be retired once the .NET standard version of it is declared stable.

patrikhellgren:
@jansenbe Great news, will check it out when available.

jansenbe:
Closing this PR as we now have PnP Framework for use in .NET Standard

PR #2790, opened 2020-10-13
Reapply PR #2596 which addresses Issue #2595

### LINKED ISSUE #2595: Provisioning Template.  Omitting Sharing Settings results is default being Disabled

michael-jensen:
This fix from @SteveClements in March was overridden in May, so I am reapplying it.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | fixes #2595 

This is from the original PR #2596 that was merged but later overridden:

As per item #2595 a simple change to `OfficeDevPnP.Core.Framework.Provisioning.Providers.Xml.Resolvers.V202002.SharingSettingsFromSchemaToModel` to return `null` when `SharingSettings` are not used in the template XML.

Currently it will return a default value which is `Disabled`

### DIFF
--- a/Core/OfficeDevPnP.Core/Framework/Provisioning/Providers/Xml/Resolvers/V202002/SharingSettingsFromSchemaToModel.cs
+++ b/Core/OfficeDevPnP.Core/Framework/Provisioning/Providers/Xml/Resolvers/V202002/SharingSettingsFromSchemaToModel.cs
@@ -39,7 +39,7 @@ namespace OfficeDevPnP.Core.Framework.Provisioning.Providers.Xml.Resolvers.V2020
                 }
             }
 
-            return (result);
+            return (null);
         }
     }
 }


gautamdsheth:
@jansenbe  - going to test this one.

PR #2801, opened 2020-11-09
Fix Layout dedection for TwoColumnRightVerticalSection and TwoColumnLeftVerticalSection

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

The TwoColumnRightVerticalSection and TwoColumnLeftVerticalSection where wrong because the compare for ColumnFactor was done with Col 1 instead of Col 0. As a result we would get 1/3 and 2/3 instead of 2/3 and 1/3 Layout with Vertical Section.


jansenbe:
Nice catch @czullu 👍

PR #2802, opened 2020-11-09
All COntent of the Field needs proper HtmlEncode not only &

czullu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

The List View Formatter Field fails during provisioning since in more complex Formatters like the new List-App provides multiple char need to be encoded in order to set the value. The Fix does a full html-encoding of the Formatter string.


jansenbe:
Thanks @czullu !

